PR #1, opened 2017-07-13
HTML report: populate "summary" bar at the bottom with stats

tripu:


tripu:
(The diff is clearer [ignoring whitespace changes](https://github.com/tidoust/reffy-reports/pull/1/files?w=1))

tidoust:
Great idea @tripu, thanks for the pull request!

I slightly adjusted the code to use CSS Flexbox to better handle widths on narrow screens. It could be interesting later on to improve that bar e.g. to distinguish between additional severities.

tripu:
Thanks, @tidoust.

Yes, much better using flexbox. I did something quick, and thought of using either flexbox or SVG in the future to make the bar more robust. That looks good!

PR #21, opened 2018-09-03
Sync IDL files with wpt upon changes to the master branch

foolip:
Fixes https://github.com/tidoust/reffy-reports/issues/20.


foolip:
I have pushed the commits as I wrote them, but this should definitely be squashed when merging.

https://travis-ci.org/tidoust/reffy-reports/builds/423920935 shows the result of running before the "limit to master" commit.

foolip:
https://github.com/web-platform-tests/wpt/pulls/autofoolip has some PRs that have been created/updated while I've been iterating on this.

foolip:
Actually, I have to undo the "get rid of stale branch cleanup" commit. Deleting branches will be the way that obsolete PRs are automatically closed, and figuring out that they are obsolete manually is wasting human effort.

foolip:
@tidoust @lukebjerring I've finished rewriting this into a single node script now. I took some inspiration from https://github.com/GoogleChromeLabs/webidl-diff/blob/lukebjerring-wpt-updater-reffy/main/update-wpt-branches.js and by putting PR creation in the same script as the branch creation I was finally able to make the branch cleanup actually work.

Please disregard the commit history, most of it is junk.

Here's the latest run, which created some PRs:
https://travis-ci.org/tidoust/reffy-reports/builds/426425542

foolip:
@zcorpan, I wonder if we might use @wpt-pr-bot to run this? It might not be me that maintains this forever, after all, and especially after taking @lukebjerring's code I don't like the look of @autofoolip everywhere.

foolip:
Thanks for review, @tidoust!

> One comment that came to mind was that this will create different PRs when we change the boilerplate of all IDL files. I suspect that's one of the reasons why you restrict the number of PRs created in a run to 5. This won't happen often in any case.

Yeah, I did that in case the boilerplate itself changes, or if the PR touching a file goes changes from "Add interfaces/foo.idl" to "Update interfaces/foo.idl" in case the foo.idl is added by someone else. A corner case, for sure.

The reason I've limited it to 5 is actually so that if all IDL files change at once, or there's a bug, we won't create 100+ PRs, and we have time to fix it after seeing the first 5 of them.

I'll go ahead and merge this to see it working. Changing the bot account will require taking the existing PRs into account anyway.

zcorpan:
@foolip sure. Can we plan that for Q1?

foolip:
@zcorpan, sure, it's pretty low prio, I'll just snooze this thread until then.

PR #23, opened 2018-09-25
Typo: Update → Add as the verb used when adding new IDL files

foolip:
This was copypasta, the whole point of the verb argument was to allow for this
distinction.

Fixes https://github.com/tidoust/reffy-reports/issues/22.


foolip:
Oops, looks like we're running Travis for PRs in a way that'll touch the real branches and PRs. Should fix that before it messes something up.

foolip:
Filed https://github.com/tidoust/reffy-reports/issues/24 and turning off Travis for PRs.

PR #26, opened 2018-10-03
[wpt-sync] Add a note when files have been recently manually updated

foolip:
Fixes https://github.com/tidoust/reffy-reports/issues/25.


foolip:
I've tested this locally with a checkout of wpt rolled back to the point where the problem described in the issue would occur, and confirmed that that branch got the extra lines, and the other commits still do not.

Showing this working with https://github.com/tidoust/reffy-reports/issues/24 would have been better, but I have the code for that dry run sitting on my laptop and didn't want to rewrite it. I'll upload that as a PR before the next largeish change here.

foolip:
It worked in "production" on this PR: https://github.com/web-platform-tests/wpt/pull/13744

However, I didn't actually spot the text, so this isn't quite prominent enough.

PR #27, opened 2018-10-03
[wpt-sync] Add --dry-run argument and use it for PRs

foolip:
Fixes https://github.com/tidoust/reffy-reports/issues/24.

### DIFF
--- a/wpt-sync/sync.js
+++ b/wpt-sync/sync.js
@@ -115,6 +115,8 @@ function listRemoteBranches(dir, remote) {
 // (if any) doesn't have the same changes as the local branch. Then remove
 // remote-only branches to close pull requests that are now no-ops.
 async function updatePullRequests(dir, localBranches, remote, remoteBranches) {
+    const dryRun = flags.get('dry-run');
+
     function git(args) {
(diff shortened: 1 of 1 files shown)


foolip:
This is **not** ready for review, I've just uploaded it so that I can work on it from all computers.

wpt-sync/sync.js
```diff
@@ -142,20 +144,29 @@ async function updatePullRequests(dir, localBranches, remote, remoteBranches) {
...
+                git(`push ${remote} ${branch}`);
+                console.log(`  Created new remote branch: ${branch}`);
+            }
         }
 
         // Next create or update a PR for the branch. Use the remote branch's
         // commit message to create the PR title/body, so that they cannot
         // get out of sync. Note that this will also update the boilerplate of
         // existing PRs even if the branch was not updated. This is intentional.
+        // FIXME: this doesn't work with dry run!
```
foolip: Problem is that no branches are pushed in dry run, and this uses the remote branches to ensure that they're in sync. Could perhaps change the ref used depending on `dryRun`.

foolip:
Replaced by https://github.com/tidoust/reffy-reports/pull/37 which works.

PR #30, opened 2019-02-08
Move package.json to the top level directory

foolip:
wpt-sync/sync.js is still the only code in this repo, but having the
whole repo be a module makes it possible to depend on it from other npm
packages.


foolip:
https://travis-ci.org/tidoust/reffy-reports/builds/490796256 (from a reverted commit) shows that the setup still works.

PR #38, opened 2019-08-23
[wpt-sync] avoid deprecated `number` arg for `octokit.pulls.update`

foolip:


foolip:
Missed in https://github.com/tidoust/reffy-reports/pull/37 because this code path isn't taken with `--dry-run`. Actual tests would be needed for this to be robust, but that seems not worth it for one code path.

foolip:
Ping @dontcallmedom @tidoust. Do either of you get updates for all new issues/PRs in this repo, or is it better if I request review manually to get past mail filters?

tidoust:
Ah, it might be better to request review manually. Not only to get past mail filters. In this particular case, I was essentially thinking that you'd merge the PR directly when ready because the update is for `wpt-sync`, which is a bit of code that you own here!

foolip:
Aha, well I'll make sure to request review explicitly in the future :)

PR #39, opened 2019-09-10
Run crawl on GitHub Actions

foolip:
Fixes https://github.com/tidoust/reffy/issues/146.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -4,13 +4,20 @@ on:
   push:
     branches:
     - master
-name: crawl
+name: update
 jobs:
-  crawl:
+  update:
     runs-on: ubuntu-18.04
     steps:
+    - name: checkout
+      uses: actions/checkout@master
     - name: checkout reffy
       uses: actions/checkout@master
+      with:
+        repository: tidoust/reffy
+        ref: master
+        fetch-depth: 1
+        path: reffy
     - name: setup node
       uses: actions/setup-node@master
       with:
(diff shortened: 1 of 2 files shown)


.github/workflows/update.yml
```diff
@@ -0,0 +1,47 @@
...
+        npm ci
+      working-directory: ../reffy
+    - name: reffy crawl
+      run: npm run all
+      working-directory: ../reffy
+    - name: copy reports
+      run: rsync -av --exclude=README.md ../reffy/reports/ ./
+    - name: git push
+      run: |
+        git config user.email auto@foolip.org
```
foolip: @tidoust what committer name and email would you like to use? It is possible to have no email, so maybe "reffy-bot" and no email?

.github/workflows/update.yml
```diff
@@ -0,0 +1,47 @@
...
+        path: reffy
+    - name: setup node
+      uses: actions/setup-node@master
+      with:
+        node-version: '10.x'
+    - name: install pandoc
+      run: sudo apt-get install pandoc
+    - name: setup reffy
+      run: |
+        echo "${{ secrets.CONFIG_JSON }}" | base64 --decode > config.json
```
foolip: This needs to be a JSON file on this form:
```
{
  "w3cApiKey": "secret"
}
```

Encoded as base64.

foolip: And that should be the only setup required for this.

foolip:
https://github.com/foolip/reffy-reports/runs/217375770 shows this running on my fork.

foolip:
https://github.com/foolip/reffy-reports/commit/0dfbb5d387afd08beea3f4f2b837f8a4dea45210 is the commit pushed.

The main difference is all the stuff in w3c-tr. Should that directory not be updated? Should it be removed instead?

.github/workflows/update.yml
```diff
@@ -0,0 +1,47 @@
...
+        npm ci
+      working-directory: ../reffy
+    - name: reffy crawl
+      run: npm run all
+      working-directory: ../reffy
+    - name: copy reports
+      run: rsync -av --exclude=README.md ../reffy/reports/ ./
```
foolip: To address a question from the previous PR, this won't delete existing files, only add.

PR #40, opened 2019-09-15
Reorganize reports in more meaninful directories

### LINKED ISSUE #33: Deprecate whatwg vs w3c views of the world?
Given the recent announcement of the MoU between WHATWG and W3C, we could probably simplify some of the code and some of the reports by deprecating the WHATWG vs W3C views of the world.

(I wonder if the W3C TR view of the world is useful / used in any way)

dontcallmedom:
whatwg and w3c moved to ed
w3c-tr moved to tr
close #33
see also https://github.com/tidoust/reffy/pull/169


foolip:
@tidoust thanks! Was a good idea to put that code in this directory, it seems!

PR #49, opened 2020-07-16
Remove crash-reporting

stephenmcgruer:
The only IDL in this was marked 'exclude' in https://github.com/WICG/crash-reporting/pull/8


stephenmcgruer:
(I have no idea if this is correct, but hopefully @foolip can tell me if/where I've made mistakes. I just searched for crash-reporting and deleted the bits that looked like they shouldn't be here).

foolip:
You ought only need to remove the IDL files, and possibly only the one in ed/. To figure out which, you'd have to find out what files reffy would still create when run, which you do, well, by running it...

Maybe @tidoust has a more satisfactory answer for how to manage this :)

tidoust:
I'm closing down this PR without merging. The discussion on WICG/reporting seems to have stalled for now. Dropping the files in webref will not be the right solution in any case. If something is needed on the extraction side, it will be an update in Reffy's crawling logic.

Issue #52, opened 2020-08-27
Migrate reffy-reports repo to w3c organization


tidoust:
Tasks to perform:

- [x] Find a way to redirect `https://tidoust.github.io/reffy-reports/` URLs to new destination? There may be external projects that actually use the `https://tidoust.github.io/reffy-reports/` URLs for now and it would be good to make sure that the migration is transparent for them. But is there a magic solution to create redirects for all existing files? Actually, is there a good solution to create redirects for non-HTML files?
=> done in https://github.com/tidoust/tidoust.github.io/commit/25f8051b0928f066cafa6f2ef155d85ee4a0c181
- [x] Migrate or create target repo under w3c organization. If people would fancy a new name, now would be a good time...
=> new [w3c/webref](https://github.com/w3c/webref) repo created
- [x] Make sure that Travis CI and job actions are enabled on the target repo
- [x] Update URL in [.travis.yml](https://github.com/tidoust/reffy-reports/blob/master/.travis.yml#L22)
=> see #53 
- [x] Update source URL for commit message in [sync.js](https://github.com/tidoust/reffy-reports/blob/master/wpt-sync/sync.js#L297)
=> see #53
- [x] Make sure that the target repo has a [CONFIG_JSON secret](https://github.com/tidoust/reffy-reports/settings/secrets)... or simply update the GitHub workflow not to require that file anymore. Reffy no longer depends on the W3C API, so it no longer needs a W3C API key!
=> dropped in https://github.com/tidoust/reffy-reports/commit/993e7ef51e925adfcb9f63d7e9c1a99f98e52a39
- [x] Create the right `GH_TOKEN` secret under Travis CI for the target repo as [done for tidoust/reffy-reports](https://travis-ci.org/github/tidoust/reffy-reports/settings) for the auto generation of PRs to WPT (requires secret from @foolip as I don't think the old secret can be copied over)

Anything missing?

Cc @dontcallmedom, @foolip

dontcallmedom:
I don't think there is a way to address the redirects (or least, no cheap way to do so, and I don't think the number of impacted people or projects to be such to be worth of using more expensive ways).

Here are some proposed names for the repo as we migrate it to w3c:
* webology
* speca-sutra
* web-isindex
* webopedia
* encycloweb
* webography
* web-atlas
* webref
* web-dictionary 
* weboscope

tidoust:
[...]

PR #53, opened 2020-09-01
[WPT sync task] Update URLs that targeted tidoust/reffy-reports

tidoust:
See #52.

@foolip, the update does not rename the branches that created, which would continue to be named `reffy-reports/xxx`. Or would you prefer to rename them as well?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 script:
   - |
     if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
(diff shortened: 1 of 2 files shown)


foolip:
Yeah, renaming the branches would be nice. It will result in recreated PRs though, so checking for the old names would be needed to avoid that.

If you don't want to deal with that it's fine to leave it as-is.

tidoust:
OK, let me merge this first so that info that appears in potential new PRs is correct, and I'll send another PR to rename the branches and still account for existing ones with the old name.

Issue #54, opened 2020-09-01
"Cannot lock ref" error on Travis CI job


tidoust:
@foolip, [latest Travis CI job](https://travis-ci.com/github/w3c/webref/builds/182272553) correctly detected the update on `webaudio.idl`, correctly created the [PR against WPT](https://github.com/web-platform-tests/wpt/pull/25330), and yet the job failed with (slightly reformatted to fit the screen):

```
{ Error: Command failed: git push fork reffy-reports/webaudio

To https://github.com/autofoolip/wpt.git

 ! [remote rejected] reffy-reports/webaudio ->
       reffy-reports/webaudio (cannot lock ref 'refs/heads/reffy-reports/webaudio': reference already exists)

error: failed to push some refs to 'https://github.com/autofoolip/wpt.git'
    at checkExecSyncError (child_process.js:629:11)
    at Object.execSync (child_process.js:666:13)
    at git (/home/travis/build/w3c/webref/wpt-sync/sync.js:141:22)
    at updatePullRequests (/home/travis/build/w3c/webref/wpt-sync/sync.js:186:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  status: 1,
  signal: null,
  output:
   [ null,
     '',
     'To https://github.com/autofoolip/wpt.git\n ! [remote rejected]
      reffy-reports/webaudio -> reffy-reports/webaudio
      (cannot lock ref \'refs/heads/reffy-reports/webaudio\': reference already exists)\n
      error: failed to push some refs to \'https://github.com/autofoolip/wpt.git\'\n' ],
  pid: 9155,
  stdout: '',
  stderr:
   'To https://github.com/autofoolip/wpt.git\n ! [remote rejected]
    reffy-reports/webaudio -> reffy-reports/webaudio
    (cannot lock ref \'refs/heads/reffy-reports/webaudio\': reference already exists)\n
    error: failed to push some refs to \'https://github.com/autofoolip/wpt.git\'\n' }

The command "if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
  node wpt-sync/sync.js --wpt-dir wpt --build-url https://travis-ci.org/w3c/webref/builds/$TRAVIS_BUILD_ID
else
  node wpt-sync/sync.js --wpt-dir wpt --dry-run
fi" exited with 1.
```

Any obvious reason for such a failure?

foolip:
[...]

PR #57, opened 2020-09-01
[WPT sync task] Rename PR branches to "webref/*"

tidoust:
Branches created on Web Platform Tests are still named after `reffy-reports/*`. Following the rename to `webref`, this update changes the name of created branches to `webref/*`.

As existing PRs are identified through the name of the branch, the code still checks for and handles existing `reffy-reports/*` pull requests. See discussion in https://github.com/w3c/webref/pull/53


wpt-sync/sync.js
```diff
@@ -166,21 +167,23 @@ async function updatePullRequests(dir, localBranches, remote, remoteBranches) {
 
     console.log('Updating pull requests:');
     for (const branch of localBranches) {
-        assert(branch.startsWith('reffy-reports/'));
+        assert(branch.startsWith('webref/'));
+        const oldBranch = branch.replace(/webref/, 'reffy-reports');
```
foolip, line 171:
```suggestion
        const oldBranch = branch.replace(/^webref/, 'reffy-reports');
```

Just to be sure :)

foolip (approved):
Looks like this will work. Can it be tested before going live?

tidoust:
> Looks like this will work. Can it be tested before going live?

I was sort of hoping that you would be able to do it :)
I tested the individual git commands but haven't tried to run the script. I'll give it a try.

tidoust:
I would say that this seems to work _well enough_ in that the script correctly creates `webref/*` branches when needed and seems to be willing to run the right push commands against the wpt repository. I'm not sure how to test that all cases are handled correctly though. Do you think that we can go ahead and merge, or is there an easy way to run more thorough tests?

foolip:
@tidoust yeah, let's just land this and see what breaks :)

tidoust:
FWIW, it did break ;) There were a couple of places where the code didn't account for the fact the local and remote branches might not have the same name. Fixed in https://github.com/w3c/webref/commit/700fd9ce6548f0f861bfbe8714de53708a4c01ef

The job still fails, but with the same "Cannot lock ref" error as reported in #54 (so a priori not something related to this PR). I see that the job still manages to create PRs on WPT though...

PR #64, opened 2020-10-03
Clean up most remaining mentions of reffy-reports

foolip:
tr/idl/*.idl still mentions reffy-reports in comments.

### DIFF
--- a/.github/workflows/update-ed.yml
+++ b/.github/workflows/update-ed.yml
@@ -9,10 +9,10 @@ jobs:
   update:
     runs-on: ubuntu-18.04
     steps:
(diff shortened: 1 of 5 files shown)


tidoust:
The pull request rolls back changes made to `wpt-sync/sync.js` to check for existing branches in WPT that may still be named after `reffy-reports`. I'm fine with that but could you confirm that this is indeed intended?

foolip:
@tidoust yes, that's intentional, because something about it wasn't quite right, and the script ended up opening new PRs for every change. So now there are no PRs open using the old branches any longer. Oops :)

PR #65, opened 2020-10-07
Delete wake-lock.*

foolip:
screen-wake-lock.* were created in:
https://github.com/w3c/webref/commit/aa218922773ffe09250d76d83c50b985a3c85f92
https://github.com/w3c/webref/commit/c248c8a3802147ad5c2f3d390fcfd424463406d0

This was caused by the publishing of https://www.w3.org/TR/screen-wake-lock/


tidoust:
Thanks. I note that there are a couple of other specs that need to be deleted as well due to specs having been renamed (`web-transport` is now `webtransport` for instance). I'll do it.

Issue #63, opened 2020-09-23
Publish crawled data with fixes to NPM?


foolip:
I just did [an experiment](https://github.com/foolip/webidl-dfns) in to package up webref/ed/idl/*.idl with fixes from [a branch](https://github.com/foolip/webref/tree/mdn-bcd-updater-fixes) and publish it as https://www.npmjs.com/package/webidl-dfns.

IDL with fixes applied is needed for [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector), and I wanted to have it as a regular dependency rather than pulling it from a branch that I keep changing.

However, having done that experiment, I wonder if there would be interest in publishing webref directly to NPM? In addition to include ed/ and tr/ as-is, we would also provide fixed versions of ed/idl/* which we guarantee will parse with a given version of webidl2.js.

One could go further if one wanted to with global consistency guarantees such as:
- No duplicate definitions
- No unknown types used
- No circular inheritance

[mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector) already does some of this, but if it could be centralized, I think WPT would also be better off making use of released data with some guarantees.

tidoust:
Totally fine publishing webref to NPM! We hadn't done it until now simply because no one had asked ;)
Happy to work on it, time permitting as usual. Feel free to beat me to it!

Also OK to include additional views with IDL guarantees in that package. I suppose that requires maintaining:
- IDL conversion code (e.g. to convert `void` to `undefined` to account for changes in the WebIDL spec that may take time to propagate across specs). Reffy actually [does a tiny bit of conversion](https://github.com/tidoust/reffy/blob/master/src/cli/parse-webidl.js#L28) when it cannot parse the WebIDL, but it does not publish the converted IDL yet.
- individual IDL fixes per spec, essentially those listed in https://github.com/foolip/webref/commits/mdn-bcd-updater-fixes, as long as underlying PRs to fix the specs haven't been merged.

Code could either go in Reffy, in this repo directly (for instance as an action that runs any time a push is made), or perhaps in a separate "webgap" repo whose final goal is to disappear...

@foolip, would you like to propose some code to create the fixed version?

foolip:
We have most of the ingredients needed for this, the question is mainly how to maintain the fixes and what should be included in the NPM package. While spec updates will often break things, it would be great if significant manual work is only needed each time broken IDL is added or changed. So the ingredients of a solution might be:
[...]

PR #68, opened 2020-10-26
Set up publishing of @webref/idl with Lerna

foolip:
Lerna is a tool for managing monorepos, and while @webref/idl is the
only package here now, the packages/* structure makes sense.

Fixes https://github.com/w3c/webref/issues/63.


foolip:
@dontcallmedom does this look sensible?

.github/workflows/test.yml
```diff
@@ -0,0 +1,21 @@
...
+      matrix:
+        node-version: ['10', '12', '14']
+    name: Node.js ${{ matrix.node-version }}
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node-version }}
+    - run: npm ci
+    - run: npx lerna bootstrap
```
saschanaz, line 20: FYI I heard that this NPM 7 now can manage monorepo: https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/

foolip: It looks like https://github.com/npm/rfcs/blob/latest/implemented/0026-workspaces.md is the feature in question.

It looks like that'll ship with Node.js 15, although it is possible to `npm install -g npm@7` and get it on older Nodes.js, I did it with Node.js 12.

@dontcallmedom would you be OK with requiring npm 7?

foolip: I'm having some trouble getting workspaces to work, and I wonder if it really has the features we want. In particular, I can't find an equivalent to `lerna publish`, and I'm not sure if `npm publish` will support publishing a new version only if there are changes.

So I'm leaning towards trying Lerna first, and migrating later if others using Lerna successfully do so.

foolip: https://github.com/w3c/webref/commit/faab9f616a8b8b230e58b4066c0190aa49c1a398 is what I tried.

dontcallmedom: using the default (even if new) npm stuff would have been my preference, but it's still immature or an ill-fit, Lerna is what @tidoust and I had discussed when we had looked into this. In particular, I assume there won't be a user impact if we later change from Lerna to npm, so starting with Lerna sounds quite reasonable.

saschanaz:
Can we also publish CSS things? I can try it as a separate PR though.

.github/workflows/test.yml
saschanaz: Yeah, the `npm publish` currently seems workspaces-unaware and there is no immediate plan to fix it: https://github.com/npm/rfcs/pull/117/files#diff-f2d6f6a359bc25897c95aa08add740198c6b7743b5f2f5a9456535c5d7244fbdR71

Personally I think it's unusable without the proper publish command 🤔

packages/idl/index.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  async text() {
+    const text = await fs.readFile(this.path, 'utf8');
+    return text;
+  }
+
+  async parse() {
+    const text = await this.text();
+    return WebIDL2.parse(text);
```
saschanaz, line 20: I feel this is somewhat overkill without real uses inside webref, could we just provide the text and let the users do the parsing part in whatever way they want?

.github/workflows/test.yml
foolip: Lerna it is then :)

packages/idl/index.js
foolip: Note that webidl2 is a peer dependency, and that ensuring that a specific version of webidl2 is able to parse all of the IDL in this package is the main value of a separate package. Are you skeptical of some of that as well, or just saying that the `parse` and `parseAll` helpers here aren't that useful?

saschanaz: Ah, my confusion, totally okay if it's a pear dependency.

foolip: Oh, alright then :)

foolip:
Ping @dontcallmedom for a sensibility review.

What's missing here is still the mechanism by which the IDL files will be fixed to be consistent. Options I've considered:
- A branch with changes which we keep rebasing, so we always release this package from that branch.
- A set of patches (as files) which are applied after copying ed/idl/* into packages/idl/.
- A branch which we treat as a big patch and apply after copying ed/idl/* into packages/idl/.

The last option is perhaps my favorite one, but it's a bit iffy that the repo wouldn't work as a shallow checkout, but depends on its own history.

dontcallmedom:
my own initial idea was to maintain the patches as files in the main branch, so that the content of the package be self-contained. What are the advantages of using a dedicated branch?

Whichever way is picked, failing visibly when the patch no longer applies will be important.

saschanaz:
The approach I'm using on TSJS is to simply ignore the failing upstreams and maintain my own IDL. Is there a reason to keep both failing and patched ones?

foolip:
@dontcallmedom, @tidoust and I met and discussed this in a VC today. The main outstanding issue is how to maintain the fixes, as discussed in https://github.com/w3c/webref/pull/68#issuecomment-724753797. Out of the options listed there, maintaining a bunch of patches as files in the repo is the easiest to get started with, so let's go with that.

foolip:
@dontcallmedom @tidoust this is ready for review now. I can't really test if the published package will look as hoped, but if we publish 1.0.0 we'll quickly know :)

@saschanaz only the fixed IDL will be published, if anyone enjoys broken IDL there's always ed/idl/*.idl in this repo :)

packages/idl/README.md
```diff
@@ -0,0 +1,43 @@
+# Web IDL definitions of the web platform
+
+This package contains Web IDL scraped from the latest versions of web platform specifications in [webref](https://github.com/w3c/webref), with fixes applied to ensure all definitions can be parsed and are internally consistent.
+
+# API
+
+The async `listAll()` method resolves with an object where they keys are spec shortnames, and the values are objects with async `text()` and `parse()` methods to get the raw text and the result of parsing that with `WebIDL2.parse`, respectively. Example:
```
dontcallmedom, line 7:
```suggestion
The async `listAll()` method resolves with an object where the keys are spec shortnames, and the values are objects with async `text()` and `parse()` methods to get the raw text and the result of parsing that with `WebIDL2.parse`, respectively. Example:
```

ed/idlpatches/trusted-types.idl.patch
```diff
@@ -0,0 +1,65 @@
+From a2a36c82e56b8956f32710c0b5d76db2e68d3544 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Philip=20J=C3=A4genstedt?= <philip@foolip.org>
+Date: Wed, 10 Feb 2021 13:24:47 +0100
+Subject: [PATCH] Remove trusted-types.idl overloads
+
+https://github.com/w3c/webappsec-trusted-types/issues/283
```
dontcallmedom, line 6: Looking at the issue, it sounds like this patch should no longer be needed.

Does this PR include a mechanism to detect patches that have become redundant?

test/idl/all.js
```diff
@@ -0,0 +1,24 @@
+const assert = require('assert');
+
+const idl = require('@webref/idl');
+
+describe('@webidl/idl module', () => {
+  it('listAll', async () => {
+    const files = await idl.listAll();
+    assert.strictEqual(Object.keys(files).length, 198);
```
dontcallmedom, line 8: I'm not sure I understand the value of hardcoding that number?

ed/idlpatches/trusted-types.idl.patch
foolip: For some reason the IDL I tried to get excluded is still there. I noticed it today but didn't look into it because it will likely take time to reach webref anyway.

If it had been removed this patch would no longer apply. Generally, the way we'll notice that something needs fixing is when patches no longer apply or when the tests start failing. I haven't written instructions for how to deal with that yet, I thought I'd learn by doing 😀

dontcallmedom:
I've been able to publish the 1.0.0 of the package using this pull request: https://www.npmjs.com/package/@webref/idl  🎉

@tidoust will provide a JS rewrite of prepare.sh (so that it can be used on Windows)

tidoust:
> @tidoust will provide a JS rewrite of prepare.sh (so that it can be used on Windows)

That's not needed, after all. The problem is that the default (and recommended) Git config on Windows makes it convert LF line endings to CRLF in text files, and bash does not like that. I'll add a `.gitattributes` file to tell Git to keep LF line endings for `.sh` files.

That said, with that fix, I note I still get the following error when I run the `prepare.sh` script for the SRI patch that should delete the file:
```
patching file SRI.idl
Reversed (or previously applied) patch detected!  Assume -R? [n]
```

Anyone knows what may explain this?

foolip:
@tidoust SRI.idl should indeed be deleted, and using the version of `patch` on my system I don't get such a complaint. But I guess that's not universal. We could instead just `rm` it in `prepare.sh`.

tidoust:
Same cause, different consequences...

[`patch`](https://man7.org/linux/man-pages/man1/patch.1.html) applies some heuristic to convert CRLF line endings to LF on Windows machines and so patch no longer matches. That problem is not restricted to the SRI patch, it just happens to be the first one to crash the script on my machine. If I remove the patch, other patches fail too with a more explicit "different line endings" error message.

The `--binary` option disables that behavior and solves the issue on my end (and should not have any impact on POSIX systems).

foolip:
It looks like we've found a situation where just a branch would have been easier to deal with, but oh well. Let's add in `--binary` and hope there's not a long list of issues like this waiting for us.

Since 1.0.0 was already published, it would be great to land this. I've discovered some issues when trying to integrate this into mdn-bcd-collector, but I'd like the original state from which 1.0.0 was published to be in the repo.

tidoust:
> Since 1.0.0 was already published, it would be great to land this

Can we just drop the hardcoded number of IDL files (`198`) in the test before we merge? Or is there a rationale to have that number hardcoded?

test/idl/all.js
foolip: I just wanted to make sure something fails if due to some mishap there are no files at all found, because I think all the other tests would still pass. How about I just check that `length` isn't 9?

dontcallmedom:
I think this is now ready to merge - @tidoust do you confirm?

Thanks a lot @foolip for driving this forward!

foolip:
I fixed a few more small things. Please squash and keep only the first commit message, because I made no effort :)

PR #69, opened 2020-11-02
Drop obsolete CSS specs, Worklets, and unversioned WebXR module specs

tidoust:
Several specifications have been removed from browser-specs, see:
https://github.com/w3c/browser-specs/pull/174
https://github.com/w3c/browser-specs/pull/179
https://github.com/w3c/browser-specs/pull/181

This update drops related extracts from webref.


tidoust:
Note that I believe that Travis CI's job fails because the PR comes from a fork, not because the updates would break it.

dontcallmedom:
ah thanks for the clarification, I hadn't understood it that way :)

dontcallmedom (approved):

PR #71, opened 2020-11-12
Set up semi-automatic removal of outdated files

### LINKED ISSUE #66: Automatically delete files that have no matching specs in browser-specs
via a pull request

dontcallmedom:
close #66

see #72 for what this currently identifies as outdated

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -1,0 +1,34 @@
+Name: Clean up abandonned files
+
+on:
+  schedule:
+    - cron: '30 0 * * 1'
+  push:
+    branches:
+    - master
+jobs:
+  update:
+    runs-on: ubuntu-18.04
+    steps:
+    - name: Checkout webref
+      uses: actions/checkout@master
+      with:
+        path: webref
+    - name: Setup node.js
+      uses: actions/setup-node@master
+      with:
+        node-version: 10.x
+    - name: Run clean up
+      run: node tools/clean-abandonned-files.js
+    - uses: peter-evans/create-pull-request@v2
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        title: Remove files no longer referenced
+        commit-message: "Remove files no longer referenced"
+        body: |
+          This removes previously extracted files that are no longer referenced in browser-specs
+        assignees: tidoust, dontcallmedom
+        draft: true
+        branch: cleanup
+        branch-suffix: timestamp
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -1,4 +1,4 @@
-Name: Clean up abandonned files
+Name: Clean up abandoned files
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,34 @@
+Name: Clean up abandonned files
```
tidoust, line 1:
```suggestion
Name: Clean up abandoned files
```

tools/clean-abandonned-files.js
```diff
@@ -0,0 +1,26 @@
...
+    const subdir = path.split("/")[1];
+    if (!ed.results.find(spec => spec[subdir] === subdir + "/" + filename)) {
+      fs.unlinkSync(path + "/" + filename);
+    }
+  }
+}
+
+for (let dir of subdirs) {
+  checkDir("ed/" + dir);
+  checkDir("tr/" + dir);
```
tidoust, line 25: What is the rationale behind comparing everything to `ed/index.json`? I would have expected the check on `tr/` to use `tr/index.json`, where the info could perhaps be slightly different (be it only because the TR crawl is only done once per week.

dontcallmedom: the rationale is me being  distracted :)

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,34 @@
+Name: Clean up abandoned files
+
+on:
+  schedule:
+    - cron: '30 0 * * 1'
+  push:
+    branches:
+    - master
```
tidoust, line 8: If the action runs when a push is made to `master`, then I don't see any reason to schedule it as a cron job.

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,34 @@
...
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      with:
+        title: Remove files no longer referenced
+        commit-message: "Remove files no longer referenced"
+        body: |
+          This removes previously extracted files that are no longer referenced in browser-specs
+        assignees: tidoust, dontcallmedom
+        draft: true
+        branch: cleanup
+        branch-suffix: timestamp
```
tidoust, line 34: If I understand the action correctly, the `branch-suffix` setting will create a new PR each time, instead of reusing an existing one. That means one new PR created every 6 hours until we merge the changes...

"Default" behavior (when `branch-suffix` is not defined) seems preferable:
https://github.com/peter-evans/create-pull-request#action-behaviour
(not sure if `delete-branch` should be set)

dontcallmedom: it will only create a pull request if there are changes, which I think would be OK if we use the cron job (leaving us a week to react), but you're right it wouldn't be if we use on-push.

(using the default behavior won't work - it means once we close the first pull request the script generates, it won't ever be opened again IIUC)

.github/workflows/cleanup.yml
dontcallmedom: right - I think keeping the cron-job only is probably best to make this more manageable

tidoust (approved):

PR #88, opened 2021-02-10
Adjust cleanup job to also look at IDL names folders

tidoust:
The "idlnames" and "idlnameparsed" folders have recently been added, see:
https://github.com/w3c/reffy/pull/489#issuecomment-775859515

They contain files per IDL Name, which need to be dropped when the IDL names no longer appear in any of the crawled specs.

Also adjust jobs execution schedules to have the cleanup job run after the weekly tr crawl, see:
https://github.com/w3c/webref/pull/86#pullrequestreview-580112062

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -2,7 +2,7 @@ name: Clean up abandoned files
 
 on:
   schedule:
-    - cron: '30 0 * * 1'
+    - cron: '0 2 * * 1'
   workflow_dispatch:
 jobs:
   update:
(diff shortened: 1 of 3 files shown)


dontcallmedom (approved):

tidoust:
Note the build error should be ignored. We should not run the webref-to-wpt job on pull requests...

tidoust:
> Note the build error should be ignored. We should not run the webref-to-wpt job on pull requests...

Actually, that's a different permission issue that has started to appear a week ago. I'll look into it. Not related to this update in any case.

PR #90, opened 2021-02-11
Fix broken links on index page

lahmatiy:


tidoust (approved):
Thanks!

(Tests fail for some unrelated reason, ignoring...)

PR #93, opened 2021-02-12
Test for conflicts when flattening IDL and adapt patches to pass

foolip:


foolip:
The Travis CI failure can be ignored. It shouldn't be running on PRs at all, but I can't be bothered to fix that, I'd like to replace the wpt IDL syncing mechanism entirely.

tidoust (commented):
Tests look good.

One thing that I believe we should fix: the second and third `it` tests depend on the first, since it is the one that sets `dfns` (and now `partials` and `includes`).

It is not too uncommon to run a single test during debugging, using the [grep option of mocha](https://mochajs.org/#command-line-usage). For instance `mocha --recursive -g "inheritance"` will match and run the second test in that file, which will happily pass no matter what... because `dfns` would then be empty.

In other words, I believe we should reconstruct `dfns` at the beginning of each test, or set it once and for all in a `before` function.

We can land this and address this afterwards though.

foolip:
@tidoust I have prepared a fix for that, and I'll land this to be able to send that PR. I'll also release 1.0.1.

PR #94, opened 2021-02-12
Switch Lerna to independent mode

foolip:
https://github.com/lerna/lerna#independent-mode

The intention is for a @webref/css package to be versioned independently
of @webref/idl. In the meantime, this avoids two places for the same
version.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.0.0"
+  "version": "independent"
 }


foolip:
There's a consequence of this which is worth pointing out. In the default mode, all packages get the same version, and `npx lerna version patch` would create a tag called `v1.0.1`. With independent mode, each package would get its own tag, so we'd get `@webref/idl@1.0.1` instead.

@dontcallmedom @tidoust do you think we should tag the releases at all? If so do you care if there are more tags in total? There is something to be said for the scoped tags, since it's not webref itself which will be at `v1.0.1`.

foolip:
Regarding tagging, I checked what commit could be retroactively tagged with `@webref/idl@1.0.0`, but it was published from https://github.com/w3c/webref/pull/68 and `master` changed before it was merged, so there is no commit which could be tagged. That's fine, just noting it so nobody else needs to spend time figuring that out.

foolip:
Regarding tagging, I went ahead and retroactively tagged [`@webref/idl@1.0.0`](https://github.com/w3c/webref/releases/tag/%40webref%2Fidl%401.0.0) on the branch which used to release. I can delete it again if you'd rather not have any tags for releases.

tidoust (approved):
Makes sense to me to have tags per package release.

PR #95, opened 2021-02-12
Split out setup to allow a subset of consistency tests to run

foolip:
The last testing (merging) still modifies the definitions, so these
tests still can't be run in any order. That would require parsing all of
the Web IDL for each test, which seems excessive.

### DIFF
--- a/test/idl/consistency.js
+++ b/test/idl/consistency.js
@@ -186,6 +186,7 @@ describe('Web IDL consistency', () => {
     }
   });
 
+  // This test should remain the last one as it slightly modifies objects in dfns in place.
   it('merging in partials/mixins', () => {
     const merged = merge(dfns, partials, includes);
   });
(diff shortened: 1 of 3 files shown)


test/idl/consistency.js
```diff
@@ -141,26 +162,31 @@ describe('Web IDL consistency', () => {
...
         assert(parent, `${dfn.name} inherits from missing type: ${dfn.inheritance}`);
-        assert.strictEqual(dfn.type, parent.type, `${dfn.name} inherits from wrong type: ${parent.type}`);
+        assert.equal(dfn.type, parent.type, `${dfn.name} inherits from wrong type: ${parent.type}`);
       }
     }
   });
 
-  it('flattening partials/mixins', () => {
-    flatten(dfns, partials, includes);
+  it('merging in partials/mixins', () => {
```
tidoust, line 189: I suggest to add a note for the future that the test should remain last because `mergeMembers` modifies the definitions in place, which could potentially afftect tests that we may add later on. Something like:

```suggestion
  // This test should remain the last one as it slightly modifies objects in dfns in place.
  it('merging in partials/mixins', () => {
```

tidoust (approved):

PR #98, opened 2021-02-15
Remove ignore rule no longer needed

foolip:


tidoust (approved):
Fixed in the spec by https://github.com/w3c/webrtc-insertable-streams/pull/52

PR #99, opened 2021-02-16
Rewrite prepare script in JS

foolip:
This is to make it easier to apply other fixes programatically,
like replacing [NoInterfaceObject] with [LegacyNoInterfaceObject].


packages/prepare.js
```diff
@@ -0,0 +1,47 @@
...
+  // The patches are against ed/idl/ and can be applied there using `git am`,
+  // but apply them in packages/idl/ instead using `git apply`.
+  // (The --binary option disables the heuristic for transforming CRLF line
+  // endings into LF line endings on Windows machines)
+  const patchFiles = await fs.readdir(patchDir);
+  for (const file of patchFiles) {
+    if (file.endsWith('.patch')) {
+      const patch = path.join(patchDir, file);
+      console.log(`Applying ${file}`);
+      await execFile('git', ['apply', '--binary', '--directory=packages/idl', '-p3', patch], {
```
tidoust, line 37: The `--binary` option seems useless on Windows machines when `git am` is used.

Actually, looking at the [documentation for `git am`](https://git-scm.com/docs/git-apply/2.29.0#Documentation/git-apply.txt---binary), it is now just a no-op, so can be dropped (along with the comment that explains why it's there).

tidoust (approved):
I confirm that runs out of the box on Windows.
It does not hurt but I note that, with that change, we can also remove the `.gitattributes` file.

packages/prepare.js
foolip: Great, I've done that cleanup now.

tidoust (approved):

PR #101, opened 2021-02-17
Bump webidl2.js to fix a spurious validation in background-fetch

foolip:
See https://github.com/w3c/webidl2.js/issues/545

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7240,9 +7240,9 @@
       "dev": true
     },
     "webidl2": {
-      "version": "23.13.0",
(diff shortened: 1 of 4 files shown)


tidoust (approved):
Not directly related, but this PR makes me wonder whether dependabot will "see" the `package.json` in `packages/idl` when it suggests to bump the version of the webidl2 parser.

If not, we should add a test that makes sure that the webidl2 version in the main `package.json` file is always the same as the one in `packages/idl/package.json`.

foolip:
I'm pretty sure Dependabot isn't going to update `packages/idl/packages.json`, so indeed we should test for that.

foolip:
I've sent https://github.com/w3c/webref/pull/103

PR #102, opened 2021-02-17
Fix the remaining spec-specific IDL validation errors:

foolip:


tidoust (approved):
LGTM. Made me discover the WebIDL rule that mandates a final dictionary parameter with no required members to be optional. The rule makes sense. It is also very easy to miss, so a good example of why running the WebIDL validator is useful.

PR #105, opened 2021-02-18
Drop content-index.idl and web-codecs.idl patches

foolip:
Both have been applied to the specs.


tidoust (approved):
Feel free to self-approve and merge patch drops directly. I'm happy to review but that adds delay and should be useless 99.99% of the time...

foolip:
@tidoust will do, thanks!

PR #106, opened 2021-02-18
Validate that all IDL types are defined somewhere

foolip:


test/idl/consistency.js
```diff
@@ -186,6 +186,79 @@ describe('Web IDL consistency', () => {
...
+      for (const [key, value] of Object.entries(node)) {
+        if (key === 'idlType' && typeof value === 'string') {
+          usedTypes.add(value);
+        } else if (typeof value === 'object' && value !== null) {
+          pending.push(value);
+        }
+      }
+    }
+
+    const knownTypes = new Set([
```
tidoust, line 210: @dontcallmedom I note that the WebIDL parser in Reffy has a similar list to avoid flagging well-known types as dependencies:
https://github.com/w3c/reffy/blob/master/src/cli/parse-webidl.js#L352

Comparing the two lists (leaving aside `void` and types defined by other specs), I note the following differences:

- `ArrayBufferView` in Reffy's WebIDL parser list but not here
- `BufferSource` in Reffy's WebIDL parser list but not here
- `DOMException` in Reffy's WebIDL parser list but not here
- `DOMTimeStamp` in Reffy's WebIDL parser list but not here
- `Error` in Reffy's WebIDL parser list but not here
- `Function` in Reffy's WebIDL parser list but not here
- `RegExp` in Reffy's WebIDL parser list but not here
- `VoidFunction` in Reffy's WebIDL parser list but not here
- `symbol` here but not in Reffy's WebIDL parser list

Most of them are from [section 4 in WebIDL](https://heycam.github.io/webidl/#common) and can probably be removed from Reffy's WebIDL parser list because they get defined as "real" WebIDL structures.

I'm not sure why `Error` and `RegExp` are in Reffy's WebIDL parser list.

Last one is `symbol`, which seems correct but does not appear in any spec that defines some WebIDL.

I don't think that any change is needed here. Obviously, "DRY" could be an idea but then lists should not have to change often so probably not worth the hassle.

tidoust (approved):

test/idl/consistency.js
tidoust: I created w3c/reffy#516 to possibly refresh the list in Reffy.

PR #108, opened 2021-02-19
Exclude svg-paths.idl from @webref/idl package

foolip:
The patch for SVG.idl is adjusted to include its definition of
SVGPathElement, which doesn't have any members.

See https://github.com/w3c/svgwg/issues/818


tidoust (approved):
Fine with this. Perhaps we could actually drop the spec from browser-specs until progress resumes. Tracked in https://github.com/w3c/browser-specs/issues/247

Issue #111, opened 2021-02-25
Publish CSS bits as @webref/css


foolip:
As a follow-up to https://github.com/w3c/webref/issues/63 we should also have a `@webref/css` package.

@saschanaz needs it for https://github.com/saschanaz/types-web/issues/66, and I'm also still using webref directly in https://github.com/foolip/mdn-bcd-collector because I get CSS properties from it.

What should be in such a package? Is it simply the JSON files in ed/css/?

saschanaz:
Autogenerating IDL per [CSSOM](https://drafts.csswg.org/cssom/#ref-for-css-property-to-idl-attribute) could be a good additional touch.

foolip:
Do you think that should go in this package, or in `@webref/idl`?

saschanaz:
Hmm, maybe it should go to the IDL package.

tidoust:
For linking purpose, autogeneration of IDL per CSSOM is tracked in #51.

saschanaz:
I wonder whether synthesizing AST for IDL would make more sense than generating IDL text, since in that case I think it fits better in `/css` as `/idl` is about IDL text.

tidoust:
The direction we've chosen so far is more to export raw material and provide some parsing guarantees. That's not perfect. We still need to use a JSON structure for CSS extracts because that is the raw-est option, and the guarantees somewhat impose to use a JS runtime, whereas data could be useful in other runtimes. I would still prefer to stick to that for the time being.

If I understand [CSSOM](https://drafts.csswg.org/cssom/#ref-for-css-property-to-idl-attribute) correctly, each property essentially generates one, two, or three attributes on the `CSSStyleDeclaration` interface. I'm wondering whether the following solution could fit be a good middle-ground:

1. In CSS extracts included in the CSS package, add a `styleDeclaration` key to property definitions that contains an array with the IDL attribute names that the property generates. For instance, leaving aside other keys:
   ```json
    {
      "properties": {
        "align-content": {
          "styleDeclaration": ["align-content", "alignContent"]
        }
      }
    }
    ```
2. Create an additional `ed/idl/cssom-generated.idl` file that gets included in the IDL package, and that contains a large `partial interface CSSStyleDeclaration` block with all the CSS properties. Or are there scenarios where it would be useful to have the CSSOM generated IDL per spec?

The information in the CSS extracts would tell where specific attributes in the additional IDL file come from. It would also allow one to easily generate the IDL text (and AST) for a given CSS property if needed.

[...]

PR #115, opened 2021-03-03
Add logic to create @webref/css package

tidoust:
Code follows the same structure as for the `@webref/idl` package. Tests make sure that all CSS fragments can be parsed with the Value Definition Syntax parser of CSSTree.

Note that this is not yet the case, there are 10 cases with syntax errors that require the creation of dedicated patches, not included in this pull request.

Actually, I realize that I don't know how to generate patch files. @foolip, what's the best way to do that?

I kept the logic in the `index.js` file simple and in particular it does not have any logic to actually parse the CSS fragments. I just put an example in the README. Let me know if you believe the `parseAll` function should also create `parsedValue` properties.

Related discussion in #111, also in Reffy in https://github.com/w3c/reffy/issues/494


tidoust:
@foolip, I believe that's now ready to merge.

I completed the pull request with CSS patches. Some of them are needed because the parser does not support certain constructs that could nevertheless be legit, see discussion in https://github.com/w3c/reffy/issues/494#issuecomment-790713119

I would address the possible addition of CSSOM generated IDL, discussed in https://github.com/w3c/webref/issues/111#issuecomment-790528301, separately.

foolip:
@tidoust I'm out this week so please feel free to merge and release the first version without waiting for me!

ed/csspatches/css-tables.json.patch
```diff
@@ -0,0 +1,26 @@
+From 093265b42f910145693bef023dc76fe2480311a7 Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Thu, 4 Mar 2021 15:50:25 +0100
+Subject: [PATCH] Drop zero-width spaces
+
+https://github.com/w3c/csswg-drafts/pull/4275
```
dontcallmedom, line 6: since that PR was merged and reverted, it would probably best to have another reference (issue or pull request)

ed/csspatches/fill-stroke.json.patch
```diff
@@ -0,0 +1,27 @@
+From 9be95c0f7f61ad2ffd296a9346c552ae05b3f519 Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Thu, 4 Mar 2021 17:00:07 +0100
+Subject: [PATCH] Separate stacked combinators
+
+The parser does not support stacked combinators, see discussion in:
+https://github.com/w3c/reffy/issues/494#issuecomment-790713119
```
dontcallmedom, line 7: I'm a bit reluctant to integrate patches that target the parser limitations rather than spec bugs;  or more specifically, I don't like the idea of packaging/releasing wrongly-fixed spec data

tidoust: I'm reluctant too but what alternative do you suggest while this gets sorted out?
Drop the `value` fields? Drop the properties? Drop the files?

dontcallmedom: patch the parser :)

ed/csspatches/css-tables.json.patch
tidoust: I prepared a PR and updated the patch: https://github.com/w3c/csswg-drafts/pull/6077

ed/csspatches/fill-stroke.json.patch
tidoust: I don't think I'll have the bandwidth to look into that in the next few days, but please do :)

foolip: Could we drop this patch and initially not guarantee that everything is possible to parse?

dontcallmedom: having started to look into patching the parser, I agree that this sounds like the best path forward to get something already useful out sooner rather than later

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+const css = require('@webref/css');
+
+const cssValues = [
+  { type: 'property', prop: 'properties', value: 'value' },
+  { type: 'extended property', prop: 'properties', value: 'newValue' },
+  { type: 'descriptor', prop: 'descriptors', value: 'value' },
+  { type: 'value space', prop: 'valuespaces', value: 'value' }
+];
+
+css.parseAll().then(all => {
```
foolip, line 13: `test/idl/validate.js` call `run()` and when I wrote that I worried about what a second test requiring async setup should do. This seems to work, but I don't understand why. Adding an artificial delay to `css.parseAll()`, the tests still don't run until this has run. I don't understand how that works.

Since it works I guess that's good, but if someone does understand how this works, I'd like to know :)

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+  for (const [shortname, json] of Object.entries(all)) {
+    describe(`The ${shortname} entry in @webidl/css`, () => {
+      for (const { type, prop, value } of cssValues) {
+        if (!json[prop]) {
```
foolip, line 17: A typo in `prop` could lead to no tests running here. Since the JSON files always have all of these properties, it works to just remove this bit, and have more confidence in the tests as a result.

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+
+const cssValues = [
+  { type: 'property', prop: 'properties', value: 'value' },
+  { type: 'extended property', prop: 'properties', value: 'newValue' },
+  { type: 'descriptor', prop: 'descriptors', value: 'value' },
+  { type: 'value space', prop: 'valuespaces', value: 'value' }
+];
+
+css.parseAll().then(all => {
+  for (const [shortname, json] of Object.entries(all)) {
```
foolip, line 14:
```suggestion
  for (const [shortname, data] of Object.entries(all)) {
```

I usually try to call only strings (or bytes) JSON, and pick some other name for the parsed value.

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+    describe(`The ${shortname} entry in @webidl/css`, () => {
+      for (const { type, prop, value } of cssValues) {
+        if (!json[prop]) {
+          continue;
+        }
+        for (const [name, desc] of Object.entries(json[prop])) {
```
foolip, line 20:
```suggestion
        for (const [name, desc] of Object.entries(data[prop])) {
```

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
+const assert = require('assert').strict;
+const { definitionSyntax } = require('css-tree');
+
+const css = require('@webref/css');
+
+const cssValues = [
+  { type: 'property', prop: 'properties', value: 'value' },
+  { type: 'extended property', prop: 'properties', value: 'newValue' },
```
foolip, line 8: What is "newValue"? It looks to me like that doesn't appear in the input JSON files, so is this accidentally dead code?

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+
+css.parseAll().then(all => {
+  for (const [shortname, json] of Object.entries(all)) {
+    describe(`The ${shortname} entry in @webidl/css`, () => {
+      for (const { type, prop, value } of cssValues) {
+        if (!json[prop]) {
+          continue;
+        }
+        for (const [name, desc] of Object.entries(json[prop])) {
+          if (!desc[value]) {
```
foolip, line 21: I've checked that this can actually happen, but I'm not sure when or if it's intentional. See above question.

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+            continue;
+          }
+          it(`defines a valid ${type} "${name}"`, () => {
+            const ast = definitionSyntax.parse(desc[value]);
```
foolip, line 25: I guess not throwing an exception is passing. `assert.doesNotThrow` would make that more explicit. If you do that, and add an assertion message including `type`, `name` and `desc[value]`, you might be able to have just a single test per shortname here and still make failures easy to diagnose. It's not very important, but I found the very large number of tests a lot to scroll through :)

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
...
+            assert(ast.type);
```
foolip, line 26: Is there anything more that can be asserted about what CSSTree returns?

test/css/validate.js
```diff
@@ -0,0 +1,32 @@
+const assert = require('assert').strict;
+const { definitionSyntax } = require('css-tree');
```
foolip, line 2: I'd suggest renaming this test to `parse.js` or maybe `css-tree.js` since it's not doing validation akin to `WebIDL2.validate`.

test/css/all.js
```diff
@@ -0,0 +1,25 @@
...
+          `invalid shortname: ${shortname}`);
+    }
+  });
+
+  it('parseAll', async () => {
+    const all = await css.parseAll();
+    assert(Object.keys(all).length > 0);
+    for (const desc of Object.values(all)) {
+      assert(desc);
+      assert(desc.spec);
```
foolip, line 21: I spotted this in the output as well. Since it's data from browser-specs, and the same convenience isn't included in `@webref/idl`, should we actually hide this and not make it part of the API? If it's needed, it would be nice to have a common way of getting this extra information for both `@webref/css` and `@webref/idl`.

packages/prepare.js
```diff
@@ -4,38 +4,57 @@ const util = require('util');
...
-  // rm dstDir/*.idl
-  const dstFiles = await fs.readdir(dstDir);
-  for (const file of dstFiles) {
-    if (file.endsWith('.idl')) {
-      await fs.unlink(path.join(dstDir, file));
+  for (const { name, srcDir, dstDir, patchDir, fileExt } of packages) {
+    // rm dstDir/*.${fileExt}
+    const dstFiles = await fs.readdir(dstDir);
+    for (const file of dstFiles) {
+      if (file.endsWith('.' + fileExt) && (file !== 'package.json')) {
```
foolip, line 31:
```suggestion
      if (file.endsWith(`.${fileExt}`) && file !== 'package.json') {
```

Although it would actually be simpler if `fileExt` included the leading period.

packages/prepare.js
```diff
@@ -4,38 +4,57 @@ const util = require('util');
...
-    if (file.endsWith('.patch')) {
-      const patch = path.join(patchDir, file);
-      console.log(`Applying ${file}`);
-      await execFile('git', ['apply', '--directory=packages/idl', '-p3', patch], {
-        cwd: rootDir
-      });
+    // The patches are against srcDir and can be applied there using `git am`,
+    // but apply them in dstDir instead using `git apply`.
+    // See ed/idlpatches/README.md for how to maintain these patches.
+    if (patchDir) {
```
foolip, line 47: Making this unconditional seems fine, unless you think we'll eventually get to no patches reliably for at least some package?

packages/prepare.js
```diff
@@ -4,38 +4,57 @@ const util = require('util');
...
+      const patchFiles = await fs.readdir(patchDir);
+      for (const file of patchFiles) {
+        if (file.endsWith('.patch')) {
+          const patch = path.join(patchDir, file);
+          console.log(`Applying ${file}`);
```
foolip, line 52:
```suggestion
          console.log(`Applying ${patch}`);
```

Otherwise it won't be easy to tell if it's a CSS or IDL patch being applied, their names are the same.

packages/css/README.md
```diff
@@ -0,0 +1,53 @@
...
+for (const [shortname, json] of Object.entries(parsedFiles)) {
+  if (json.properties) {
+    for (const [name, desc] of json.properties) {
+      if (desc.value) {
+        const ast = definitionSyntax.parse(desc.value);
+        // do something with the ast
+      }
+    }
+  }
+}
```
foolip, line 46:
```suggestion
const parsedFiles = await css.parseAll();
for (const [shortname, data] of Object.entries(parsedFiles)) {
  for (const [name, desc] of Object.entries(data.properties)) {
    if (desc.value) {
      const ast = definitionSyntax.parse(desc.value);
      // do something with the ast
    }
  }
}
```
```

**Review by foolip — COMMENTED** [review]
Looking forward to seeing this published :)

File: test/css/validate.js (hunk 7 above)
Comment (foolip) [reply test/css/validate.js:8]: Aha, I think it should be "newValues", which is produced for things like https://drafts.csswg.org/css-flexbox-1/#flex-containers as a kind of partial value definition.

File: packages/css/index.js [hunk 17]
```diff
@@ -0,0 +1,37 @@
...
+    const text = await fs.readFile(this.path, 'utf8');
+    return JSON.parse(text);
+  }
+}
+
+async function listAll() {
+  const all = {};
+  const files = await fs.readdir(__dirname);
+  for (const f of files) {
+    if (f.endsWith('.json') && (f !== 'package.json')) {
```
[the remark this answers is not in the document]
tidoust: I find the former form more readable but I understand I'm in the minority here ;)

packages/css/README.md
```diff
@@ -0,0 +1,53 @@
+# CSS definitions of the web platform
+
+This package contains CSS property definitions scraped from the latest versions of web platform specifications in [webref](https://github.com/w3c/webref), with fixes applied to ensure all CSS value definitions can be parsed and are internally consistent.
```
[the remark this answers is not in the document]
tidoust: That makes sense. There are tests that we could add over time to make sure that we have all the CSS properties and values that we need. Right now, there are values that are not used anywhere (e.g. because I believe that we do not yet extract @rules) and there are likely missing values (e.g. because these values are defined in prose).

packages/css/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+# API
+
+The async `listAll()` method resolves with an object where the keys are spec shortnames, and the values are objects with an async `json()` method to get the contents of the file as a JSON object. Example:
```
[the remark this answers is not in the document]
tidoust: OK, I'll update to only have a `listAll()` function that does what `parseAll()` currently does.

test/css/validate.js
tidoust: I had noticed the `run()` in your tests and actually did not understand what that `run()` function was referring to, so I dropped it and that did not seem to have any consequence. Thinking about it some more, I don't really understand why this works either. I'll look into it.

test/css/validate.js
tidoust: I was actually thinking that we might want to drop keys from the CSS extracts when values are empty sets, to avoid useless constructs such as `"descriptors": {}`. However, I see how guaranteeing presence also provides additional confidence that the code generates the right keys in the end. I'll update the test accordingly.

test/css/validate.js
tidoust: Argh, indeed, that was a typo, it should have been `newValues`!

```suggestion
  { type: 'extended property', prop: 'properties', value: 'newValues' },
```

test/css/validate.js
tidoust: There are several definitions for which the value is defined in the prose. In such cases, all we have is a `prose` key, for example, see result for [`<timeline-name`](https://github.com/w3c/webref/blob/master/ed/css/css-animations-2.json#L41).

Also, `newValues` is used when the definition extends a base definition.

There should always be a `value`, `newValues` or `prose` key. I'll add a test to assert that this is indeed the case.

packages/css/README.md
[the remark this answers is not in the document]
foolip: Indeed. I started writing some code for https://github.com/foolip/mdn-bcd-collector/issues/705 and it quickly became clear that what's extracted from CSS specs isn't consistent. There are duplicate definitions and missing definitions.

test/css/validate.js
foolip: This is using https://mochajs.org/#delayed-root-suite, but that documentation alone really provides no clues about why it doesn't break when the promises resolve in the wrong order...

test/css/validate.js
foolip: Great, thank you!

packages/prepare.js
foolip: I see, if we should ever get down to zero patches even temporarily, there would be no directory, and `await fs.readdir(patchDir)` would throw. I guess we could add a `.gitignore` to always keep the directory around, or just handle that case. Adding and removing the `patchDir` setting above would work, but the failure mode would be less clear.

test/css/validate.js
tidoust: And not calling `run()` does have consequences: the tests won't run if they are run in isolation, e.g. with `mocha --recursive --delay test/css/`. Which makes complete sense. I'll add the missing `.then(run)`.

foolip:
Trying to write some code to solve https://github.com/foolip/mdn-bcd-collector/issues/705, I wrote what could become consistency tests:
https://github.com/w3c/webref/commit/9ea71c102728848fadabeb42a2a6ad3647b962bf

The output of that and the many manually added valuespaces could form a burndown list of what needs to be fixed in specs and possibly in Reffy to extract something that's consistent. A good example is that https://drafts.csswg.org/css-shapes-1/#typedef-basic-shape doesn't define the syntax as "<inset()> | <circle()> | <ellipse()> | <polygon()> | <path()>" but merely has a list of allowed values.

test/css/validate.js
tidoust: So, that's not always the case for now. The exceptions to the rule are mostly `-webkit-` prefixed properties and legacy properties that are defined in prose but for which we currently don't extract more than the property name:
- the 62 [`-webkit-` aliases properties](https://compat.spec.whatwg.org/#css-simple-aliases) extracted from the compat spec
- the 3 [legacy group properties](https://drafts.csswg.org/css-align/#gap-legacy) in css-align
- [`-webkit-line-clamp`](https://drafts.csswg.org/css-overflow/#propdef--webkit-line-clamp) in css-overflow
- [`-webkit-user-select`](https://drafts.csswg.org/css-ui/#propdef--webkit-user-select) in css-ui
- [`stop-color`](https://svgwg.org/svg2-draft/pservers.html#StopColorProperty) and [`stop-opacity`](https://svgwg.org/svg2-draft/pservers.html#StopOpacityProperty) in SVG

I'd be tempted to treat that separately after this PR lands.

packages/prepare.js
tidoust: The patch folders have a README file, so they should never be empty in practice.

packages/css/README.md
```diff
@@ -0,0 +1,39 @@
...
+```js
+const css = require('@webref/css');
+
+const parsedFiles = await css.listAll();
+for (const [shortname, json] of Object.entries(parsedFiles)) {
+  // do something with the json object
+}
+```
+
+CSS fragments that appear in the JSON objects, in other words the contents of the `properties[].value`, `properties[].newValue`, `descriptors[].value` and `valuespaces[].value` properties can be parsed with the [CSSTree Value Definition Syntax parser](https://github.com/csstree/csstree/blob/master/docs/definition-syntax.md#value-definition-syntax). Example:
```
foolip, line 18:
```suggestion
CSS fragments that appear in the JSON objects, in other words the contents of the `properties[].value`, `properties[].newValues`, `descriptors[].value` and `valuespaces[].value` properties can be parsed with the [CSSTree Value Definition Syntax parser](https://github.com/csstree/csstree/blob/master/docs/definition-syntax.md#value-definition-syntax). Example:
```

test/css/validate.js
tidoust: Probably. I haven't looked into that for now. I propose to address further improvements afterwards.

test/css/all.js
tidoust: I note that, in a way, the same info appears in `@webref/idl`, it's just hidden in comments because we're using a text format, e.g.:
```
// Source: Accelerometer (https://w3c.github.io/accelerometer/)
```

I would keep the info there, mostly because it's useful when you manually browse one of these files and want to understand where the info comes from and/or check the source spec.

It could still be useful to have an `index.js` file in the browser-specs package that exports a function that takes a series or spec shortname as input and returns the current specification that matches the shortname. Or did you have something else in mind?

test/css/validate.js
tidoust: I created a couple of issues to track that: #126 and https://github.com/w3c/reffy/issues/538

packages/prepare.js
tidoust: I dropped the conditional.

test/css/validate.js
tidoust: Tracked in #127.

ed/csspatches/fill-stroke.json.patch
tidoust: I dropped the patches. Tests ignore the problematic entries for the time being.

tidoust:
> Trying to write some code to solve [foolip/mdn-bcd-collector#705](https://github.com/foolip/mdn-bcd-collector/issues/705), I wrote what could become consistency tests:
> [9ea71c1](https://github.com/w3c/webref/commit/9ea71c102728848fadabeb42a2a6ad3647b962bf)
> 
> The output of that and the many manually added valuespaces could form a burndown list of what needs to be fixed in specs and possibly in Reffy to extract something that's consistent. A good example is that https://drafts.csswg.org/css-shapes-1/#typedef-basic-shape doesn't define the syntax as "<inset()> | <circle()> | <ellipse()> | <polygon()> | <path()>" but merely has a list of allowed values.

We did not maintain that list, which is now likely outdated, but we had done a similar exercise in Reffy with a list of [missing value definitions](https://github.com/w3c/reffy/blob/master/src/specs/missing-css-rules.json).

tidoust:
I believe I integrated the feedback received. @foolip, how does that look now?

packages/css/README.md
```diff
@@ -0,0 +1,43 @@
+# CSS definitions of the web platform
+
+This package contains CSS property definitions scraped from the latest versions of web platform specifications in [webref](https://github.com/w3c/webref), with fixes applied to ensure ([almost](#guarantees)) all CSS value definitions can be parsed with [CSSTree](https://github.com/csstree/csstree).
```
foolip, line 3: Do we still have this guarantee? If not this and the guarantees section below needs updating.

packages/css/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+# API
+
+The async `listAll()` method resolves with an object where the keys are spec shortnames, and the values are the result of parsing the CSS files as JSON. Example:
```
foolip, line 7:
```suggestion
The async `listAll()` method resolves with an object where the keys are spec shortnames, and the values are the data for that spec. Example:
```

The fact that JSON is used internally in the package need not show up here I think.

packages/css/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```js
+const css = require('@webref/css');
+
+const parsedFiles = await css.listAll();
+for (const [shortname, json] of Object.entries(parsedFiles)) {
```
foolip, line 13:
```suggestion
for (const [shortname, data] of Object.entries(parsedFiles)) {
```

packages/css/README.md
```diff
@@ -0,0 +1,43 @@
...
+  // do something with the json object
+}
+```
+
+CSS fragments that appear in the JSON objects, in other words the contents of the `properties[].value`, `properties[].newValues`, `descriptors[].value` and `valuespaces[].value` properties can be parsed with the [CSSTree Value Definition Syntax parser](https://github.com/csstree/csstree/blob/master/docs/definition-syntax.md#value-definition-syntax). Example:
```
foolip, line 18:
```suggestion
CSS fragments that appear in the objects, in other words the contents of the `properties[].value`, `properties[].newValues`, `descriptors[].value` and `valuespaces[].value` properties can be parsed with the [CSSTree Value Definition Syntax parser](https://github.com/csstree/csstree/blob/master/docs/definition-syntax.md#value-definition-syntax). Example:
```

test/css/all.js
foolip: A util like that in browser-specs is what I had in mind. But let's not dabble with that now. We can release 2.0.0 any time we want to change things.

packages/css/README.md
tidoust: That is why I added the "(almost)" with a link to the guarantees section, which explains that there remains a handful of exceptions in practice.

test/css/all.js
tidoust: Filed https://github.com/w3c/browser-specs/issues/256.

PR #117, opened 2021-03-08
Document how to maintain the IDL patches

foolip:


tidoust (approved):
Many thanks for this!

I'll add a README to the patch folder of the upcoming CSS package that links to this.

PR #124, opened 2021-03-09
Remove svg-paths.idl.patch

foolip:
The svg-paths.idl file is removed.


foolip:
Tests fail, but those are additional failures, not caused by this.

Issue #123, opened 2021-03-09
Release @webref/idl every Monday


foolip:
Per https://github.com/w3c/browser-specs/pull/252#issuecomment-793884959 it would be great to arrange to @webref/idl releases on Mondays.

@dontcallmedom @tidoust what would a process for this look like? If we have steps written down for how to do the release, could we take turns doing it, or even automate? Do you automatically publishing any other NPM packages?

foolip:
Releasing on Thursday would actually be great and is what I originally wanted to suggest before https://github.com/w3c/browser-specs/pull/252#issuecomment-793884959. This is because BCD is released every Thursday, and it makes sense to depend on BCD and `@webref/idl` together, which is what mdn-bcd-collector does. The update PRs for both should then come on Friday, which would be convenient.

If we don't both with automating releases for now, what steps should a human follow to determine if a release is warranted, and which part of the version should be bumped?

tidoust:
Ideally, a semi-automated release process could be:
1. A job runs every Thursday and creates a PR that updates the version in `packages/idl/package.json` and details the expected diff in the body of the PR description. Default would probably be to increase the `patch` part of the version number. If the job fails, e.g. because a patch is no longer valid, it should rather create an issue.
2. One of us reviews the PR, checks the diff, checks the need to release a minor / major version instead.
3. When the PR is merged, another job reacts to the version increment in `packages/idl/package.json`, and creates the release and publishes the package.

I'm not sure to what extent lerna is needed with that approach. As far as I can tell, it would only take care of tagging the release and publishing the package, which may also be done with a couple of Git and npm commands.

foolip:
@tidoust that sounds pretty good to me.

A non-trivial bit is creating a diff of the changes, since the previous state is only in an NPM package. Do we install a copy of the most recent package, and diff against that?

foolip:
I discovered in https://github.com/w3c/webref/pull/129#issuecomment-797374120 that publishing using `npx lerna publish --force-publish` no longer works when we have two packages, as both must then be updated, so I had to do it manually. So we'll have no use for Lerna at all I think.

saschanaz:
Is there any remaining work item here?

foolip:
No, this is done. Thanks @tidoust!

PR #133, opened 2021-03-15
Add jobs to semi-automate publication of @webref/idl package

tidoust:
This roughly implements the logic discussed in https://github.com/w3c/webref/issues/123#issuecomment-796650510:

1. The `prepare-idl-release` job runs on Thursday morning. It compares the latest contents of the repo with the published `@webref/idl` package on NPM, and creates a PR that bumps the version of the package in `packages/idl/package.json`, and exposes the diff of the IDL.

2. The `release-idl` job runs whenever an update to `packages/idl/package.json` gets pushed to the default branch. It publishes a new @webref/idl package to npm and creates a new @webref/idl release on GitHub. The job is a no op if the version number in package.json does not differ from the one published on npm.

If all goes well, merging the PR created by the first job will trigger the second job.

The update also moves the cleanup job to Wednesday morning instead of Monday.

If a patch no longer applies when the first job runs, it will fail. It won't create an issue in such cases, but repo admins will receive a "job failed" notification.

To be entirely clear, I have no idea whether these jobs are going to work as expected :) That said, at worst, they should simply do nothing and we should be able to iterate progressively to get them to do something useful.

If things work, the logic can easily be duplicated for the `@webref/css` package.

### DIFF
--- a/.github/workflows/prepare-idl-release.yml
+++ b/.github/workflows/prepare-idl-release.yml
@@ -1,0 +1,89 @@
+# Create a PR that bumps the version of packages/idl/package.json when changes
+# are detected compared with the latest version of the package published on NPM.
+#
+# Typical weekly schedule:
+# - New browser-specs release on Monday,
+# - Integration of new version of browser-specs in Reffy on Tuesday
+# - Cleanup job on Wednesday
+# - (This job) New @webref/idl package release on Thursday
+
+name: Prepare new @webref/idl release PR if needed
+
+on:
+  schedule:
+    - cron: '0 4 * * 4'
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Setup node.js
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14.x
+
+    # Checkout webref and prepare new release in a separate folder so that we
+    # may also install the current version of @webref/idl for comparison.
+    - name: Checkout webref
+      uses: actions/checkout@v2
+      with:
+        path: webref
+
+    - name: Prepare new release
+      run: |
+        npm ci
+        npm run prepare
+      working-directory: webref
+
+    - name: Test new release
+      run: npm test
+      working-directory: webref
+
+    - name: Install current @webref/idl package for comparison
+      run: |
+        mkdir idl
+        cd idl
+        npm install @webref/idl
+
+    - name: Compute diff between new release and current package
+      # For details on syntax to use to export a multiline string to an
+      # environment variable, see:
+      # https://github.com/actions/toolkit/blob/main/docs/commands.md#set-an-environment-variable
+      # The diff does not take the package.json file into account because
+      # "npm install" adds properties that start with "_" to that file which do
+      # not exist in the repo version of the file.
+      run: |
+        echo 'IDL_DIFF<<EOF' >> $GITHUB_ENV
+        diff idl/node_modules/@webref/idl webref/packages/idl --ignore-trailing-space --exclude=package.json >> $GITHUB_ENV
+        echo 'EOF' >> $GITHUB_ENV
+
+    - name: Bump version number in packages/idl/package.json
+      if: ${{ env.IDL_DIFF }}
+      run: echo "RELEASE_VERSION=$(npm version patch)" >> $GITHUB_ENV
+      working-directory: webref/packages/idl
+
+    - name: Create pull request
+      uses: peter-evans/create-pull-request@v3
+      if: ${{ env.IDL_DIFF }}
+      env:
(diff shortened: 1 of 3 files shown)


.github/workflows/prepare-idl-release.yml
```diff
@@ -0,0 +1,89 @@
...
+    - name: Compute diff between new release and current package
+      # For details on syntax to use to export a multiline string to an
+      # environment variable, see:
+      # https://github.com/actions/toolkit/blob/main/docs/commands.md#set-an-environment-variable
+      # The diff does not take the package.json file into account because
+      # "npm install" adds properties that start with "_" to that file which do
+      # not exist in the repo version of the file.
+      run: |
+        echo 'IDL_DIFF<<EOF' >> $GITHUB_ENV
+        diff idl/node_modules/@webref/idl webref/packages/idl --ignore-trailing-space --exclude=package.json >> $GITHUB_ENV
```
foolip, line 58: Is the path here right? With no `working-directory` here, what's the implicit working directory? I would have guessed it's the default location for a checkout, not the parent and this seems to assume.

.github/workflows/release-idl.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Publish npm package
+      id: publish
+      uses: JS-DevTools/npm-publish@v1
+      with:
+        token: ${{ secrets.NPM_TOKEN }}
+        package: ./packages/idl/package.json
+        access: public
+
+    - name: Create GitHub release
```
foolip, line 46: I didn't create a GitHub release for any of the earliest releases, what happens if there is none, or if this last step sometimes fails?

foolip (approved):
LGTM with questions.

dontcallmedom (approved):

.github/workflows/prepare-idl-release.yml
tidoust: I do not really know but isn't that the same thing as in the `update-ed` job?
https://github.com/w3c/webref/blob/master/.github/workflows/update-ed.yml#L37

foolip: Perhaps, I didn't compare it to anything. Let's try and see if it works?

.github/workflows/release-idl.yml
tidoust: I'm confused by the Releases page on GitHub:
https://github.com/w3c/webref/releases

I'm the one who created a release for the CSS package. The page also lists IDL package releases, but perhaps they are just tags created with `git tag`? If you did not create the tags yourself, then Lerna most likely did.

I'll push an update to the PR to simply create a tag for now. It seems good practice to tag the commit associated with a particular npm package. We don't need something more fancy for now.

tidoust: Done in last commit. If these steps fail, then that means the npm package will not be directly associated with a commit. We should be notified of the job's failure and can always add the tag manually afterwards.

tidoust:
FYI, the preparation job fails at the diff step. It may be due to the "echo" syntax (which I don't necessarily understand but copied over from the [doc](https://github.com/actions/toolkit/blob/main/docs/commands.md#set-an-environment-variable)). Or perhaps simply because the diff command fails (e.g. because the paths are incorrect). I'll make a couple of updates directly on the default branch to isolate the error, then watch, rinse and repeat...


The reported error is:
```
> Run echo 'IDL_DIFF<<EOF' >> $GITHUB_ENV
  echo 'IDL_DIFF<<EOF' >> $GITHUB_ENV
  diff idl/node_modules/@webref/idl webref/packages/idl --ignore-trailing-space --exclude=package.json >> $GITHUB_ENV
  echo 'EOF' >> $GITHUB_ENV
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
Error: Unable to process file command 'env' successfully.
Error: Invalid environment variable value. Matching delimiter not found 'EOF'
```

tidoust:
In practice, everything works as expected. Problem is that, when there is a diff, the `diff` command outputs the diff but also exits with a `1` exit code, which gets intepreted as failure. What is the easiest way to ignore the exit code?

dontcallmedom:
I don't know if it's the most recommendable way, but `diff foo bar || echo -n` is guaranteed not fail

tidoust:
That seems to work, Job created https://github.com/w3c/webref/pull/138, which seems correct.

Remaining verification steps:
- merge that PR to check the release job.
- make sure that the prepare job does not create a PR when there are no diffs.

tidoust:
Yeah, NPM publication works! See https://www.npmjs.com/package/@webref/idl/v/1.0.9
Creation of a tag failed because the committer name needs to be specified, see https://github.com/w3c/webref/pull/139

PR #148, opened 2021-03-23
Rebase the uievents.idl patch (parts of it applied)

foolip:
https://github.com/w3c/uievents/pull/297 should fix what remains by
fixing the spec build/deploy.


foolip:
The tests still fail because of other things, but the patches now apply cleanly.

PR #149, opened 2021-03-23
Add bigint as a known Web IDL type

foolip:
Also patch the IDL where it was found.


foolip:
The remaining failures are now because of specs that have been renamed.

dontcallmedom (approved):

PR #151, opened 2021-03-24
Remove patches for uievents and webrtc-encoded-transform

saschanaz:
The patches are now redundant.


tidoust (approved):
Thanks!

FWIW, obsolete files get removed semi-automatically. The ones in this PR got removed by https://github.com/w3c/webref/pull/152.

PR #159, opened 2021-04-01
Add temporary patch for WebTransport

tidoust:
The patch rolls back the IDL to what it was before the introduction of mixins with inheritance.

### DIFF
--- a/ed/idlpatches/webtransport.idl.patch
+++ b/ed/idlpatches/webtransport.idl.patch
@@ -1,0 +1,53 @@
+From 4e7d7acc3a55b774b4a1836e6374c43642cb6e9a Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Thu, 1 Apr 2021 10:38:13 +0200
+Subject: [PATCH] [PATCH] Rollback to previous version to fix mixin inheritance
+
+https://github.com/w3c/webtransport/pull/241
+https://github.com/w3c/webtransport/issues/239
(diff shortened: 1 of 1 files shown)


tidoust:
(An alternative would be for the patch to apply the changes in https://github.com/w3c/webtransport/pull/241 but the update has not been approved yet)

dontcallmedom (approved):
https://github.com/w3c/webtransport/pull/241 hasn't been approved, but is semantically equivalent to something that was approved before, so would probably be reasonable to pre-include. But in practice, I think reverting to the previous state is the easiest path to release this week

foolip:
I've filed https://github.com/w3c/webtransport/issues/243 about this.

ed/idlpatches/webtransport.idl.patch
```diff
@@ -0,0 +1,53 @@
...
+ [ Exposed=(Window,Worker) ]
+-interface mixin OutgoingStream : WritableStream /* of Uint8Array */ {
++interface mixin OutgoingStream {
++  readonly attribute WritableStream writable;
+   readonly attribute Promise<StreamAbortInfo> writingAborted;
+   undefined abortWriting(optional StreamAbortInfo abortInfo = {});
+ };
+ 
+ dictionary StreamAbortInfo {
+-  [EnforceRange] octet errorCode = 0;
```
foolip, line 28: We don't really need to revert this change, but doing so is harmless.

foolip (approved):

PR #168, opened 2021-04-13
Patch remaining cases of NoInterfaceObject

foolip:
NoInterfaceObject.patch touches multiple files and is thus unusual, but
it works just the same.

### DIFF
--- a/ed/idlpatches/webgl-extensions.idl.patch
+++ b/ed/idlpatches/webgl-extensions.idl.patch
@@ -1,0 +1,152 @@
+From a61b9321f310cfce1ca0f79a0c0f65a0d7a6ab52 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Philip=20J=C3=A4genstedt?= <philip@foolip.org>
+Date: Tue, 13 Apr 2021 09:44:13 +0200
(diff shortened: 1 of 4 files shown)


ed/idlpatches/NoInterfaceObject.patch
```diff
@@ -0,0 +1,151 @@
+From b4ddf97e8642974a2b4a7f6910309dd950c44710 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Philip=20J=C3=A4genstedt?= <philip@foolip.org>
+Date: Tue, 13 Apr 2021 09:44:13 +0200
+Subject: [PATCH] Replace NoInterfaceObject with LegacyNoInterfaceObject
+
```
tidoust, line 5: It would be tedious to list individual pull requests, but perhaps a link to check which ones remain?

```suggestion

https://github.com/KhronosGroup/WebGL/pulls?q=is%3Apr+is%3Aopen+nointerfaceobject

```

foolip: Good idea! I see those also add `Exposed=Window`, so it would be better if my patch here did that too to avoid future conflicts.

foolip: OK, done!

tidoust (approved):

PR #171, opened 2021-04-14
Patch remaining cases of void

foolip:
All remaining cases have open pull requests or issues.

### DIFF
--- a/ed/idlpatches/SVG-void.idl.patch
+++ b/ed/idlpatches/SVG-void.idl.patch
@@ -1,0 +1,227 @@
+From 36cbbb25c6a82b7b5ca9d694e3db56a8fc466d1d Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Philip=20J=C3=A4genstedt?= <philip@foolip.org>
+Date: Wed, 14 Apr 2021 09:11:50 +0200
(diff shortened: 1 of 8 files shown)


foolip:
This will conflict with https://github.com/w3c/webref/pull/168, but resolving the conflicts will be trivial.

ed/idlpatches/SVG-void.idl.patch
```diff
@@ -0,0 +1,185 @@
+From a8fbc3ca92b35de70d2639bc4fbe902b97c30c17 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Philip=20J=C3=A4genstedt?= <philip@foolip.org>
+Date: Wed, 14 Apr 2021 09:11:50 +0200
+Subject: [PATCH] Fix void in SVG.idl
```
foolip, line 4: Note that this is `ed/idlpatches/SVG-void.idl.patch` instead of merging with `ed/idlpatches/SVG.idl.patch`, on the expectation that this patch can soon be dropped while the existing one cannot.

foolip: Actually I'm going to fold `svg-animations.idl.patch` into this, since it's from the same spec PR.

PR #172, opened 2021-04-14
Patch remaining cases of missing [Exposed]

foolip:


foolip:
This depends on https://github.com/w3c/webref/pull/168 and there will be some conflicts to resolve, but the hard work is done.

tidoust:
@foolip, not sure if you want to separate patches for `filter-effects` or combine them, I let you address the conflict!

foolip:
@tidoust I ended up splitting the SVG patches a bit differently since I couldn't get the `svg-animations.idl` patches to apply if they were separate, since they touch lines close to each other.

tidoust (approved):

PR #174, opened 2021-04-14
Delete large and outdated crawl.json files

tidoust:
Reffy no longer creates the `ed/crawl.json` and `tr/crawl.json` files, see:
https://github.com/w3c/reffy/pull/569

Note final goal is to strike the file from history entirely afterwards to get back to a reasonable repo size:
https://github.com/w3c/reffy/issues/385#issuecomment-816701140

By definition, this means losing the history of the `crawl.json` files. The `index.json` file was introduced in August 2020, so it will be hard to extract a consistent crawl result before that date, but I doubt we'll ever need that:
https://github.com/w3c/reffy/commit/0b7a5c5ae3076b8dc82840e3c879a16939466ca6

As far as I can tell, pruning the files from history leaves the commits intact, they will just no longer contain anything about `crawl.json`. For instance, this commit should essentially become a no-op... I'll give it a try on a fork first.


tidoust:
Tests fail because one of the patches is no longer valid. Not related to this PR.

dontcallmedom (approved):

PR #181, opened 2021-04-15
Update webidl2 to v24.0.1

saschanaz:


tidoust (approved):

foolip:
@tidoust should we consider this a breaking change for webref as well and bump the major version, or how should we deal with an incompatible `peerDependency` change?

packages/idl/package.json
```diff
@@ -12,6 +12,6 @@
   "license": "MIT",
   "main": "index.js",
   "peerDependencies": {
-    "webidl2": "^23.13.1"
+    "webidl2": "^24.0.1"
```
saschanaz, line 15: I didn't bother here but this can also be a range that covers both 23 and 24.

foolip: I think that's the right call. To continue to have confidence in that we'd need to test both versions, and that's not worth it.

PR #183, opened 2021-04-16
Release @webref/idl@v2.0.0

foolip:
This major release is because the webidl2 peerDependency was bumped:
https://github.com/w3c/webref/pull/181

See https://github.com/semver/semver/issues/502 for precedent.

### DIFF
--- a/packages/idl/package.json
+++ b/packages/idl/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@webref/idl",
   "description": "Web IDL definitions of the web platform",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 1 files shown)


tidoust:
Tests should no longer fail now that #185 was merged.
I'll let @dontcallmedom handle this. I don't feel strongly about it but we had come to the opposite conclusion, in other words that a major version bump seemed unwarranted.

dontcallmedom (approved):
the reasoning for the opposite conclusion is that the breaking change in webidl2.js itself isn't really breaking; but given that this seems to be a well documented pattern in semver, it's probably simplest to go with the flow.

dontcallmedom:
can you rebase @foolip ?

foolip:
Tests now pass, merging.

PR #188, opened 2021-04-21
Print a relative path when applying CSS/IDL patches

foolip:
This makes it easier to copy-paste the path in the output, and it's also
less verbose + consistent locally and in CI.


dontcallmedom (approved):
LGTM - will merge after rebasing on top of #189 to confirm tests no longer fail

PR #192, opened 2021-04-21
Rebase the MathML patch

foolip:
https://github.com/mathml-refresh/mathml-core/pull/31 was merged.


foolip:
The tests are still going to fail, but I've confirmed locally that the patch now applies.

PR #196, opened 2021-04-22
Validate all IDL together to catch additional issues

foolip:
Because DOMMatrix2DInit is defined in geometry.idl, when validating just
SVG.idl in isolation it's not known that DOMMatrix2DInit has no required
members, so the error about missing default {} is missed.


foolip:
I noticed this because in https://github.com/foolip/mdn-bcd-collector/ I added validation of all spec IDL + custom IDL as a set, and a few of the errors weren't in the custom IDL...

dontcallmedom (approved):

PR #207, opened 2021-05-06
Add IDL patches for Web Animations 2 and WebCodecs

tidoust:


ed/idlpatches/web-animations-2.idl.patch
```diff
@@ -0,0 +1,78 @@
+From 77438b7fa65e8f54ee68c9d224e55a1552bbb3e9 Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Thu, 6 May 2021 10:48:57 +0200
+Subject: [PATCH] [PATCH] void to undefined, default value for optional param
+
+https://github.com/w3c/csswg-drafts/pull/6271
+
+Patch also drops duplicate `fillMode` enum definition. The spec is a delta spec
+and re-defines that enum because it updates the meaning of one of its values.
+We'll likely need to keep the patch around for this as long as the spec remains
```
dontcallmedom, line 10: Would it make sense to keep that more permanent aspect in a separate patch?

dontcallmedom (approved):
LGTM, with a non-blocking suggestion

ed/idlpatches/web-animations-2.idl.patch
tidoust: I'm worried that maintaining two patches for the same file may create conflicts (some of the `void` changes are around the `fillMode` declaration), so I lazily prefer to redo the patch once the PR has been merged.

Issue #145, opened 2021-03-18
Automatic package release should start from the state of the repo that was reviewed


dontcallmedom:
as discussed in #144, there is a risk that the diff being reviewed in the PR that bumps the package version no longer match the state of the repo when that PR is merged.

The GH action that does the release should start from the state of the repo at the time of the diff.

We could for instance tag the repo as `pre-x.y.z` when generating the diff, and use that tag for the checkout in the release action.

foolip:
This would suggest actually tagging the very branch that was pushed as a PR. But then the version bumps would always be on a branch and need to be duplicated on master somehow.

One option is to bump the version right *after* a release. This is how versioning in Chromium works, the version in the tree is the version that current changes will be included in, not the version of an already cut release.

dontcallmedom:
hmm... you're right on the problem of tagging master with that approach, I had not gone fully through the steps.

I'm not sure I understand how the tag-at-release time would help if we keep the manual diff review process though - if master changes between the time of the diff & the release, there is no longer a state where master can be tagged with the version number and the expected state of the repo?

foolip:
What we'd end up doing is actually to tag a commit already on master, and then bump the version after that. The tagged commit isn't guaranteed to be the last with that version however.

What I'm suggesting doesn't work well as a PR workflow though.

Maybe release from a branch and merge that into master with a merge commit? Not sure how to enforce that though.

dontcallmedom:
I guess if we want to ensure a reliable way to retrieve the state of the repo that matches what get shipped, we probably want to have two stable branches: `master` (to be renamed `main`) which reflects the latest crawl, and something like `releases` that gets updated only with what ships in released packages and where the relevant tags would live.

saschanaz:
Would "Require branches to be up to date before merging" GitHub setting prevent the state issue?

foolip:
@saschanaz I think that would avoid the problem, yeah. If changes do happen on the main branch, some mechanism to recreate the PR is needed, and I'm not sure how to do that.

tidoust:
> Would "Require branches to be up to date before merging" GitHub setting prevent the state issue?

Where is that setting? Is it the "Require linear history" branch protection rule?


> If changes do happen on the main branch, some mechanism to recreate the PR is needed, and I'm not sure how to do that.

[...]

PR #211, opened 2021-05-08
Introduce improved release workflow

tidoust:
This updates implements the release workflow described in:
https://github.com/w3c/webref/issues/145#issuecomment-805668484

As opposed to the current release workflow, this workflow will release the version that already appears in the `package.json` files, and not the bumped version. Hence the patch bump on these files in this update.

The pre-release job takes care of creating a pre-release PR whenever needed. The job updates the pre-release PR when the diff (or the version to release) changes.

The job no longer runs on a fixed schedule, and rather runs whenever a push is made that could create a diff with the current npm package. This may mean that there will often be an open pre-release PR to review. If CSS and IDL change too much, we could perhaps only assign reviewers to the pre-release PR only once per week, instead of immediately when the PR gets created.

The second job runs when a pre-release PR gets merged, and releases the package to npm. The content of the package is taken at the main branch commit on which the pre-release PR is based.


tidoust:
Tests fail due to IDL inconsistency, not because of the changes in this PR.

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,33 @@
...
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+    - name: Setup node.js
+      uses: actions/setup-node@v2
+      with:
+        node-version: 14.x
+
+    - name: Checkout webref
+      uses: actions/checkout@v2
```
foolip, line 24: Will this check out the branch of the PR that was closed? Or a merge commit of that and main, as usually happens with PRs?

.github/workflows/release-package.yml
```diff
@@ -0,0 +1,33 @@
...
+#
+# Job does nothing if PR that was merged is not a pre-release PR.
+
+name: "Publish @webref package if needed"
+
+on:
+  pull_requests:
+    branches:
+      - 'release-*'
+    types: [closed]
```
foolip, line 11: Will this trigger is the PR is closed without merging too?

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+  // Extract released version (will be used in the body of the pre-release PR)
+  latestReleasedVersion = require(path.join(tmpFolder, "node_modules", "@webref", type, "package.json")).version;
+
+  // Diff does not take the package.json file into account because "npm install"
+  // adds properties that start with "_" to that file which do not exist in the
+  // repo version of the file.
+  // The "diff" command exits with a non zero status code when there is a diff,
+  // which would throw an exception. Final "echo" command turns that status
+  // code to 0 to avoid the exception.
+  const installedIdl = path.join(tmpFolder, "node_modules", "@webref", type);
```
foolip, line 67: Might be CSS, not necessarily IDL?

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+  let diff = execSync(
+    `diff ${installedIdl} packages/${type} --ignore-trailing-space --exclude=package.json --unified=3 || echo -n`,
```
foolip, line 69: Will this list newly added files? I noticed the latest release has new IDL files which didn't seem to be listed in the prepared PR.

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+
+  // Clean up tmp folder
+  try {
+    rimraf.sync(tmpFolder);
+  }
+  catch {
+  }
+
+  // On Windows, the "echo -n" command actually outputs "-n"
+  // Also replace tmp folder name in diff to make it possible to compare diff
```
foolip, line 80: Have you looked for a way to invoke diff that omits the leading parts of the paths in the output? Since `patch -p3` can strip them I half suspect there's a way.

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+ * @function
+ * @param {String} type Package name. One of "css" or "idl"
+ */
+async function prepareRelease(type) {
+  // Compute a reasonably unique ID
+  const uid = (new Date()).toISOString().replace(/[\-T:\.Z]/g, "");
+
+  console.log("Look for a pending pre-release PR");
+  const searchResponse = await octokit.search.issuesAndPullRequests({
+    q: `repo:${owner}/${repo} type:pr state:open 📦 Release @webref/${type}@ in:title`
```
foolip, line 105: Shouldn't the title be quoted here?

And it might be easier to just use two unique branch name patterns and find the PR by branch name.

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+  const bumpedPackageFileContents = btoa(JSON.stringify(packageFile, null, 2));
+  console.log(`- Version to release: ${version}`);
+  console.log(`- Bumped version: ${bumpedVersion}`);
+
+  console.log();
+  console.log("Get latest commit on default branch");
+  const latestCommitResponse = await octokit.repos.getCommit({
+    owner, repo, ref: "HEAD"
+  });
+  const latestCommitSha = latestCommitResponse?.data?.sha;
```
foolip, line 156: Using `?.` but still not checking if the result is null makes the code look undecided about the possibility of not finding the latest SHA.

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+
+const packageType = (process.argv[2] === "css") ? "css": "idl";
+
+prepareRelease(packageType)
+  .then(() => {
+    console.log();
+    console.log("== The end ==");
+  })
+  .catch(err => {
+    console.error(err);
```
foolip, line 304: Also `process.exit(1)`?

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+*******************************************************************************/
+const GH_TOKEN = (() => {
+  try {
+    return require("../config.json").GH_TOKEN;
+  } catch {
+    return process.env.GH_TOKEN;
+  }
+})();
+if (!GH_TOKEN) {
+  console.error("Env variable GH_TOKEN must be set to some personal access token");
```
foolip, line 287: Or config.json

tools/prepare-release.js
```diff
@@ -0,0 +1,305 @@
...
+
+🤖 This pull request was automatically created to facilitate human review of \`@webref/${type}\` changes.
+
+🧐 Please review the diff below and version numbers. If all looks good, merge this pull request to release the changes to npm.
+
+📦 Latest released \`@webref/${type}\` package was **v${latestReleasedVersion}**. Merging this pull request will release **v${version}**. Make sure that the bump is the right one for the changes.
+
+✍ If any change needs to be made before release, **do not add a commit** to this pull request. Changes should rather be handled in a separate pull request and pushed to the default branch. You may leave this pull request open in the meantime, or close it. The pre-release job will automatically update this pull request or create a new one once the updates have made their way to the default branch.
+
+🛈 Merging this pull request will bump the patch part of the version number in \`packages/${type}/package.json\` so that a new release cycle may begin.
```
foolip, line 233: Maybe explain where the bump happens since it takes some pondering to understand the model?

tools/release-package.js
```diff
@@ -0,0 +1,158 @@
...
+
+const prereleasePR = parseInt(process.argv[2], 10);
+
+releasePackage(prereleasePR)
+  .then(() => {
+    console.log();
+    console.log("== The end ==");
+  })
+  .catch(err => {
+    console.error(err);
```
foolip, line 157: Ditto

.github/workflows/release-package.yml
tidoust: Yes, hence the need for the `if: github.event.pull_request.merged == true` check 4 lines below. I could not find a direct way to trigger a worfklow on a merge action.

.github/workflows/release-package.yml
tidoust: That's unrelated to the PR. This will checkout the latest commit on the main branch. This checkout is only needed to be able to run the latest version of the `release-package.js` script.

tools/prepare-release.js
tidoust: I [checked the `diff` manual](https://www.man7.org/linux/man-pages/man1/diff.1.html) and could not find one.

tools/prepare-release.js
tidoust: Both good points. I'll change that.

I was under the impression that quotes did not exist in GitHub's search API, but [quotes can be used](https://docs.github.com/en/github/searching-for-information-on-github/understanding-the-search-syntax#use-quotation-marks-for-queries-with-whitespace) indeed.

And I thought that branch name had to be the exact branch name to match, but that is also wrong, searching for `release-idl-` or `release-css-` works fine.

tools/prepare-release.js
tidoust: I reformulated the text. I'm not fully convinced that the new text is clear, but when I try to add more text, I have the feeling that it actually becomes more confusing... Happy to take suggestions :)

In the end, what matters is to make it clear that the file changed by the PR is uninteresting for the purpose of issuing a new release. It is just there to reset the cycle after the package has been released.

tools/prepare-release.js
tidoust: Looking at the most recent pre-release PR (https://github.com/w3c/webref/pull/208), I see that it did report on newly added files:
```
[...]
Only in webref/packages/idl: sanitizer-api.idl
Only in webref/packages/idl: scheduling-apis.idl
[...]
Only in webref/packages/idl: web-animations-2.idl
```

These lines are very easy to miss though, especially since they are not coloured as the rest of the diffs. I cannot find a diff option to make them stand out a bit more though :( For instance, it would be good if there was a way to add a newline between diff outputs

dontcallmedom: we could grep "^Only in webref/packages/" into a separate output that would be presented separately from the rest of the diff?

dontcallmedom:
> The job no longer runs on a fixed schedule, and rather runs whenever a push is made that could create a diff with the current npm package. This may mean that there will often be an open pre-release PR to review.

If I understand correctly, this means we will have pre-releases pull requests opened or updated 2 or 3 times a day outside of week-ends. What's the rationale for that approach vs scheduled pre-releases? Invalidating pull request reviews on updates (which I see marked as a TODO) might be quite important to avoid merging unreviewed changes.

tidoust:
> If I understand correctly, this means we will have pre-releases pull requests opened or updated 2 or 3 times a day outside of week-ends.

The code runs on week-ends too ;)

> What's the rationale for that approach vs scheduled pre-releases?

Less release code to maintain. The same job does the creation of the PR and the management of that PR.

> Invalidating pull request reviews on updates (which I see marked as a TODO) might be quite important to avoid merging unreviewed changes.

To avoid merging unreviewed changes, I guess I would propose another approach to start with: encourage reviewers to merge the release directly after review, instead of just issuing an ok without merging. Do we really need two or more reviews for now?

My worry is more that, if we are in a situation where there always is a pre-release PR that is pending review, then reviewers are going to ignore notification messages and no one is going to review and merge the pull request. It may make sense to only request a review only once per week.

tidoust:
> merging directly is OK, but I was worried about publishing too many package updates; but maybe that's not worth worrying about.

I updated the pre-release job not to add reviewers to start with, and added a job that runs weekly on Thursday to add reviewers. Goal is to strike a good balance between releasing too often and forgetting to release when needed. We may improve on that later on.

.github/workflows/release-package.yml
tidoust: Sorry, I just realized that my previous comment is totally wrong. The [checkout action](https://github.com/actions/checkout#checkout-v2) will checkout the ref/sha that triggered the workflow. In this case, this will be the [last merge commit on the PR branch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request).

In turn, this means that possible fixes/changes made to the release script after a pre-release PR has been created would not be taken into account when the PR is merged. That seems counter-intuitive and not a good thing. I'll update the job to checkout the main branch instead.

foolip:
I won't have time to further review this this week. Sorry for the pile of comments and running, feel free to ignore anything that didn't make sense.

foolip:
> I may be missing something but how would that simplify the workflow?

The main thing I was thinking was that both the before and after state would be an npm package. But since it ends up just comparing files on the filesystem, it's not really any simpler.

tidoust:
Discussed with @dontcallmedom. A new version of the IDL package has just been published. I will now merge this. First thing that should happen is that a new PR will be created to publish an updated CSS package. And then we'll see!

PR #213, opened 2021-05-17
Add IDL patches for CSS Typed OM and Payment Request

tidoust:
- CSS Typed OM: restore previous definition of CSSColor as new definition references not yet defined CSSPercentishArray type
- Payment Request: restore PaymentAddress and AddressErrors interfaces, dropped from the spec but still referenced by "Method Payment: Basic Card" and by the Contacts API

### DIFF
--- a/ed/idlpatches/css-typed-om.idl.patch
+++ b/ed/idlpatches/css-typed-om.idl.patch
@@ -3,11 +3,7 @@ From: Francois Daoust <fd@tidoust.net>
 Date: Mon, 17 May 2021 14:21:51 +0200
 Subject: [PATCH] Restore previous definition for CSSColor
 
-New definition references not yet defined `CSSPercentishArray` structure:
-https://github.com/w3c/css-houdini-drafts/commit/8d1990684d86ad7cf1e58ba414114c8811c04ee3
-
-Property should eventually get fixed by:
-https://github.com/w3c/css-houdini-drafts/issues/1037
+https://github.com/w3c/css-houdini-drafts/issues/1041
 ---
  ed/idl/css-typed-om.idl | 7 +++----
  1 file changed, 3 insertions(+), 4 deletions(-)
--- a/ed/idlpatches/css-typed-om.idl.patch
+++ b/ed/idlpatches/css-typed-om.idl.patch
@@ -1,0 +1,35 @@
+From 19fbd9a43ddae066c864e4dbcd3493cd96e7aadb Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Mon, 17 May 2021 14:21:51 +0200
+Subject: [PATCH] Restore previous definition for CSSColor
+
+New definition references not yet defined `CSSPercentishArray` structure:
+https://github.com/w3c/css-houdini-drafts/commit/8d1990684d86ad7cf1e58ba414114c8811c04ee3
+
+Property should eventually get fixed by:
+https://github.com/w3c/css-houdini-drafts/issues/1037
+---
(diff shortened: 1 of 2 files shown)


ed/idlpatches/payment-request.idl.patch
```diff
@@ -0,0 +1,53 @@
+From d00f84ad858cc457d0e894218931ef6b28cc6d4d Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Mon, 17 May 2021 13:47:50 +0200
+Subject: [PATCH] Restore PaymentAddress and AddressErrors interface
+
+Still used in Contact API:
+https://github.com/WICG/contact-api/pull/39
+
```
dontcallmedom, line 8: wouldn't it make more sense to instead patch the Contact API and Basic Card?

My particular worry is to figure out what will trigger the removal/update of that patch.

ed/idlpatches/css-typed-om.idl.patch
```diff
@@ -0,0 +1,35 @@
+From 19fbd9a43ddae066c864e4dbcd3493cd96e7aadb Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Mon, 17 May 2021 14:21:51 +0200
+Subject: [PATCH] Restore previous definition for CSSColor
+
+New definition references not yet defined `CSSPercentishArray` structure:
+https://github.com/w3c/css-houdini-drafts/commit/8d1990684d86ad7cf1e58ba414114c8811c04ee3
+
+Property should eventually get fixed by:
+https://github.com/w3c/css-houdini-drafts/issues/1037
```
dontcallmedom, line 10: maybe, but I think it would be best if there was a separate issue to track the lack of definition for `CSSPercentishArray`

ed/idlpatches/payment-request.idl.patch
tidoust: Possibly, but whereever I put it, updates to the specs are unlikely going to trigger a merge conflict for the patch, so I suspect that we will have to track the issues referenced in the patch no matter what to determine whether the patch may be dropped.

dontcallmedom: hmm... surely, when  https://github.com/WICG/contact-api/pull/39  gets merged, this will create a conflict if we have an equivalent here?

ed/idlpatches/css-typed-om.idl.patch
tidoust: I didn't do it because the first commit message alludes to the need to define the type, and the second issue discusses changing the type further, so the group seems already well aware that the interface needs to be updated.

Also, as opposed to the other patch, any update to the interface will invalidate this patch, making is easy to detect when the patch needs to be dropped.

ed/idlpatches/payment-request.idl.patch
dontcallmedom: after discussion with @tidoust, and given the unclear future of Basic Card, reverting to the latest consistent state sounds like the better approach

dontcallmedom (approved):

ed/idlpatches/payment-request.idl.patch
tidoust: Discussed live. Generic idea is to stick to "past" and not implement updates in advance. We'll leave this as is for now.

ed/idlpatches/css-typed-om.idl.patch
dontcallmedom: issue filed at https://github.com/w3c/css-houdini-drafts/issues/1041

PR #245, opened 2021-05-28
Pre-release workflows: truncate diff when needed

### LINKED ISSUE #244: Pre-release workflows: trim PR description to 65536 characters
When a new property gets added to all files, the diff can be very large and exceed the maximum PR body accepted by GitHub (65536 character long), e.g.:
https://github.com/w3c/webref/runs/2697338218?check_suite_focus=true

Diff needs to be trimmed when that happens. Note that the comparison between successive diffs will no longer be correct as a result.

tidoust:
This update truncates the diff when it cannot fit in a PR description. The full diff gets dumped to the console instead.

Fixes #244.

### DIFF
--- a/tools/prepare-release.js
+++ b/tools/prepare-release.js
@@ -153,7 +153,7 @@ async function prepareRelease(type) {
 
   console.log();
   console.log("Compute diff between package and repo contents");
-  const diff = computeDiff(type);
+  let diff = computeDiff(type);
   console.log(`- Diff length: ${diff?.length}`);
   if (!diff) {
     if (pendingPR) {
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
LGTM - is there any motivation to the 60,000 max?

tidoust:
> 
> 
> LGTM - is there any motivation to the 60,000 max?

It's rather a lack of motivation to count the rest of the description precisely. This leaves a bit of room for extending the description if needed, and avoids having to think about off-by-1 errors ;)

dontcallmedom:
> It's rather a lack of motivation to count the rest of the description precisely. This leaves a bit of room for extending the description if needed, and avoids having to think about off-by-1 errors ;)

I guess my question was imprecise, but I'm guessing there is an actual upper bound of 65,536 characters then (I couldn't find the documentation of that limit)

tidoust:
I don't know where that limit is documented either. It appears in the error message returned by the GitHub API in the logs:

```
Validation Failed:
{
  "resource": "Issue",
  "code": "custom",
  "field": "body",
  "message": "body is too long (maximum is 65536 characters)"
}
```

PR #255, opened 2021-06-02
Add workflow that checks patch issues/PRs

### LINKED ISSUE #215: Consider job that checks status of issues referenced by patches
In some cases, patches become invalid when the spec is fixed upstream. In other cases, spec updates don't necessarily invalidate the patch. Patches reference issues and pull requests. A job could run regularly to extract them and detect whether the pull requests have been merged and issues addressed. When all issues/PR of a patch have been closed, the job could create a PR to drop the patch.

tidoust:
This creates an additional workflow that runs once per week, extracts the list of GitHub issues and PR referenced by patches, checks the status of these issues and PRs, and creates a PR that drops patches that target now closed issues and PRs.

As often, I don't necessarily guarantee that the actual workflow file does what I'd like it to do...

Fixes #215.

### DIFF
--- a/.github/workflows/clean-patches.yml
+++ b/.github/workflows/clean-patches.yml
@@ -1,0 +1,37 @@
+name: Clean patches when issues/PR are closed
+
+on:
+  schedule:
+    - cron: '30 3 * * 3'
+  workflow_dispatch:
+jobs:
+  clean:
+    runs-on: ubuntu-18.04
+    steps:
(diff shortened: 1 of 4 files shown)


dontcallmedom (approved):
I haven't reviewed the clean-patches.js code in details, but I've successfully tested it locally. very nice job, didn't expect we would see this coming in so quickly! :)

PR #271, opened 2021-06-14
Add IDL patch for WebCodecs

tidoust:
Reverts the IDL to what it was before the introduction of dictionary attribute.

### DIFF
--- a/ed/idlpatches/webcodecs.idl.patch
+++ b/ed/idlpatches/webcodecs.idl.patch
@@ -1,0 +1,104 @@
+From 7bb8d539a667ac3145e863c856944e3e862d5b6a Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Mon, 14 Jun 2021 19:01:19 +0200
+Subject: [PATCH] Revert IDL while dictionary as attribute issue gets fixed
+
+https://github.com/w3c/webcodecs/issues/278
+---
+ ed/idl/webcodecs.idl | 47 +++++++++++++++++++-------------------------
+ 1 file changed, 20 insertions(+), 27 deletions(-)
+
+diff --git a/ed/idl/webcodecs.idl b/ed/idl/webcodecs.idl
+index dbb3cc2de..16d6bd5ed 100644
+--- a/ed/idl/webcodecs.idl
++++ b/ed/idl/webcodecs.idl
+@@ -161,7 +161,6 @@ dictionary VideoEncoderConfig {
+   [EnforceRange] unsigned long displayHeight;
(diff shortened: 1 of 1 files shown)


ed/idlpatches/webcodecs.idl.patch
```diff
@@ -0,0 +1,104 @@
...
+
+diff --git a/ed/idl/webcodecs.idl b/ed/idl/webcodecs.idl
+index dbb3cc2de..16d6bd5ed 100644
+--- a/ed/idl/webcodecs.idl
++++ b/ed/idl/webcodecs.idl
+@@ -161,7 +161,6 @@ dictionary VideoEncoderConfig {
+   [EnforceRange] unsigned long displayHeight;
+   HardwareAcceleration hardwareAcceleration = "allow";
+   DOMString scalabilityMode;
+-  BitrateMode bitrateMode = "variable";
```
dontcallmedom, line 19: should we only revert the bits that are problematic? do we have a policy or precedents?

tidoust: I think I confused the previous commit for webcodecs.idl in webref and the previous commit for the source spec. There have been more than one commit that changes the IDL in the WebCodecs spec between two updates in webref in practice.

Anyway, I rolled back the rollback. The attribute is now preserved.

ed/idlpatches/webcodecs.idl.patch
```diff
@@ -0,0 +1,96 @@
...
+
+diff --git a/ed/idl/webcodecs.idl b/ed/idl/webcodecs.idl
+index dbb3cc2de..4fc4f975f 100644
+--- a/ed/idl/webcodecs.idl
++++ b/ed/idl/webcodecs.idl
+@@ -275,23 +275,22 @@ enum AudioSampleFormat {
+ [Exposed=(Window,DedicatedWorker)]
+ interface VideoFrame {
+   constructor(CanvasImageSource image, optional VideoFrameInit init = {});
+-  constructor(sequence<PlaneInit> planes,
```
dontcallmedom, line 19: my previous comment applies also to this and other bits below - do we revert only the problematic bits (dictionaries used as attributes) or all the changes that were brought in at the same time?

tidoust: My worry is that the other updates are more intertwined and were done at the same time. I don't know whether the IDL makes a lot of sense if you only apply part of the changes. I would rather err on the safe side and freeze the IDL to what it was before.

dontcallmedom: Thanks - that makes sense to me

dontcallmedom (approved):

PR #275, opened 2021-06-16
Clean patches workflow: Fix RegExp for issue extraction

tidoust:
The workflow failed to detect a second issue URL on the following line on systems that have `\n` as line separator (I had not detected that issue because my system uses `\r\n`...)

The fix is to use lookahead and lookbehind assertions in the RegExp instead of non-capturing groups, so as not to "consume" bounding characters.

This fixes the issue detected in #274.

### DIFF
--- a/tools/clean-patches.js
+++ b/tools/clean-patches.js
@@ -40,7 +40,7 @@ async function dropPatchesWhenPossible() {
   console.log();
   console.log("Extract list of issues");
   const diffStart = /^---$/m;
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
This is all `RegExp`ese to me, but I too have tested it and found it to work as expected

tidoust:
CI fails due to MathML Core files having been accidently removed. Should get fixed next time crawl runs.

Issue #269, opened 2021-06-14
SupportedType is used as argument but not declared


jodinathan:
Can't seem to find the declaration of ```SupportedType``` however, it is used in the interface DOMParser: https://github.com/w3c/webref/blob/0128e8f1ba0a91298d2ae633c8d345cb3f5bfae4/tr/idl/DOM-Parsing.idl#L9

is this expected?

dontcallmedom:
we only provide consistency guarantees on the `webref/ed` tree (where `SupportedType` has been renamed to the well-defined `DOMParserSupportedType` and the interface itself been moved to the HTML standard)

At the moment, I wouldn't recommend working on top of `webref/tr`. @tidoust maybe we should emphasize that somehow in the README?

dontcallmedom:
in the general, you'll get the highest level of guarantees from the data in the [NPM package](https://www.npmjs.com/package/@webref/idl)

tidoust:
> Can't seem to find the declaration of SupportedType [...] is it expected?

It is not fantastic but it is "expected" in the sense that the (outdated) /TR version of DOM Parsing and Serialization uses a weird `extraidl` class instead of `idl` for the declaration of the enumeration, and thus the crawler misses the definition. We could perhaps extend the crawler to also extract `pre` sections that have an `extraidl` class, but then the right fix would be to publish a new Working Draft of the spec to correctly flag the IDL section as `idl`.

> we only provide consistency guarantees on the webref/ed tree (where SupportedType has been renamed to the well-defined DOMParserSupportedType and the interface itself been moved to the HTML standard)

The `webref/ed` tree does not really provide additional consistency guarantees. It just uses the latest version of Editor's Drafts which, more often than not, are more consistent than the sometimes outdated Working Drafts published to /TR. That said, the `webref/ed` tree can also contain invalid IDL that sometimes get introduced in Editor's Drafts...

Guarantees are enforced in the NPM package (through a set of patches that get applied to the IDL before packaging)

> At the moment, I wouldn't recommend working on top of webref/tr. @tidoust maybe we should emphasize that somehow in the README?

Yes, will do.

jodinathan:
> in the general, you'll get the highest level of guarantees from the data in the [NPM package](https://www.npmjs.com/package/@webref/idl)

The point is that I am not using NodeJS and this is the github linked in the package, so I thought the packaged used these files.

I started only using the ED directory, however, I've found out that PaymentAddress is used but is only declared in the TR version.

[...]

PR #281, opened 2021-06-17
Add SVG 1.1 only interfaces to SVG patch

### LINKED ISSUE #278: Elements extraction: unknown interfaces for some SVG 1.1 elements
The switch to the /TR version of SVG 1.1 reveals a bug in the elements extraction logic: the `elements` folder contains one file per spec but the `idl` folder contains one file per spec series.

The /TR version of SVG 1.1 defines interfaces that no longer exist in SVG 2 such as [`SVGTRefElement`](https://www.w3.org/TR/SVG11/text.html#InterfaceSVGTRefElement). The corresponding IDL does not appear in webref because we only extract the IDL of the "current" spec in the series, meaning SVG 2.

Ideally, we'd have the elements and interfaces from SVG 2 + the elements and interfaces from SVG 1.1 that are not defined in SVG 2.

Possible ways to solve the problem:
1. Switch to one file per specification series for the `elements` folder (+ delta specs) as done for `css` and `idl`. The SVG 1.1 elements would then no longer be extracted, which seems wrong, but then the SVG 1.1 interfaces are *not* extracted either for the time being.
2. Exclude SVG 1.1 from the `elements` extraction.
3. Define SVG 2 as a delta spec over SVG 1.1, but that is not true in practice and will likely create IDL inconsistencies as SVG 2 redefines some of the IDL defined in SVG 1.1.
4. Add an IDL patch for `SVG.idl` that adds the interfaces defined in SVG 1.1 that are not in SVG 2. To be dropped once SVG 1.1 gets obsoleted?

tidoust:
This is a best effort attempt to add missing SVG interfaces defined in SVG 1.1 and needed by SVG 1.1 elements. This is best effort in the sense that interfaces had to be adjusted to pass IDL validation, and may not fully match the interfaces shipped in browsers.

Some parts are clumsy too, such as the dependency on `CSSValue` (defined in DOM Level 2) that is now also defined in the patch.

Fixes #278.


dontcallmedom:
hmm… with `SVGStylable` having been merged within `SVGElement` in SVG 2, we 're going to have issues when we start detecting overlapping declaration of interface members (such as `href` that was independently identified in #269); I also don't like the idea of re-adding `CSSValue` for a method that was explicitly deprecated (and probably never implemented).

I'm not sure yet how to handle this - maybe we could take a more minimal approach to the IDL (essentially reducing it to extending SVGElement)? or add an exception to the elements extraction? clearly a suboptimal situation overall… maybe we need to revisit the SVG1.1 addition and dig more into the MDN needs in this space?

tidoust:
Yes, I went through this exercice precisely to understand what would be needed but I don't like the result either :)

I would prefer to drop SVG 1.1 altogether but a more minimal approach could work as well. Essentially, this would mean dropping `SVGStylable`, `CSSValue` and `SVGExternalResourcesRequired`. I'll do that.

On top of this, if we have some clarity on elements of SVG 1.1 that we want to preserve, we could also combine the IDL patch with an elements patch that drops obsoleted elements. Do we have a list of SVG 1.1 elements to preserve (on top of those that are still defined in SVG 2)?

dontcallmedom:
> Do we have a list of SVG 1.1 elements to preserve (on top of those that are still defined in SVG 2)?

So, unless I'm wrong the list of elements is : `tref altGlyph altGlyphDef altGlyphItem  glyphRef color-profile cursor animateColor font glyph missing-glyph hkern vkern font-face font-face-src font-face-uri font-face-format font-face-name`

Looking at browser-compat-data with 
```sh
for i in tref altGlyph altGlyphDef altGlyphItem  glyphRef color-profile cursor animateColor font glyph missing-glyph hkern vkern font-face font-face-src font-face-uri font-face-format font-face-name ; do (grep version_added $i.json|grep -v "false"|grep -v "null" > /dev/null) && echo $i ; done
```

it looks like only `altGlyph`, `glyphRef`, `cursor` and `font` have documented implementations -  so maybe that's our list?

dontcallmedom (approved):

PR #285, opened 2021-06-24
Update IDL patch for Web Animations

tidoust:
Patch is still needed but IDL right before the patch got changed in the spec invalidating the previous patch.

### DIFF
--- a/ed/idlpatches/web-animations-2.idl.patch
+++ b/ed/idlpatches/web-animations-2.idl.patch
@@ -1,6 +1,6 @@
-From 02c8a7920291c53ab07942930c4234c240aff8f1 Mon Sep 17 00:00:00 2001
+From 7bcaa17bb3609a233727d8964f6bc310deecc830 Mon Sep 17 00:00:00 2001
 From: Francois Daoust <fd@tidoust.net>
-Date: Thu, 6 May 2021 15:32:18 +0200
+Date: Thu, 24 Jun 2021 10:22:36 +0200
 Subject: [PATCH] Drop duplicate `fillMode` enum definition
 
 The spec is a delta spec and re-defines that enum because it updates the meaning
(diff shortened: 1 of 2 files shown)


dontcallmedom:
The said recent changes in web-animations-2 redefine the types of existing elements in `ComputedEffectTiming` which creates overlapping IDL definitions that makes the test fail.

I guess we need to add removing these duplicates in the patch?

tidoust:
Yes :( Will do.

ed/idlpatches/web-animations-2.idl.patch
```diff
@@ -1,23 +1,31 @@
...
 --- a/ed/idl/web-animations-2.idl
 +++ b/ed/idl/web-animations-2.idl
-@@ -38,8 +38,6 @@ partial dictionary ComputedEffectTiming {
-     CSSNumberish?        progress;
+@@ -30,16 +30,6 @@ partial dictionary OptionalEffectTiming {
+     double playbackRate;
  };
  
+-partial dictionary ComputedEffectTiming {
+-    CSSNumberish         startTime;
```
dontcallmedom, line 22: sorry, I think we still need to keep `startTime` which only appears in v2 AFAICT

tidoust: True

tidoust: The type of all other properties also change in that dfn to `CSSNumberish`. I'm wondering whether it wouldn't be preferable to add a patch for Web Animations Level 1 that drops `ComputedEffectTiming`?

dontcallmedom: I wondered about it too; but then we would still need a patch to make `ComputedEffectTiming` no longer a *partial* dictionary. I don't have strong feelings (except that delta specs are a pain)

tidoust: I did that. Feeling adventurous today...

dontcallmedom (approved):

saschanaz:
Do we have an upstream tracking bug for this?

dontcallmedom:
we didn't think we could raise one given that web-animation-2 is explicitly marked as a delta spec (where this kind of overlap is to be expected); but maybe the partial declaration of the dictionary should be flagged?

ed/idlpatches/web-animations-2.idl.patch
```diff
@@ -1,21 +1,33 @@
...
 --- a/ed/idl/web-animations-2.idl
 +++ b/ed/idl/web-animations-2.idl
-@@ -33,8 +33,6 @@ partial dictionary ComputedEffectTiming {
-     double startTime;
+@@ -30,7 +30,7 @@ partial dictionary OptionalEffectTiming {
+     double playbackRate;
+ };
+ 
+-partial dictionary ComputedEffectTiming {
++dictionary ComputedEffectTiming {
```
saschanaz, line 25: This needs `: EffectTiming`, right? I don't see why not.

dontcallmedom: you're absolutely right, we missed that completely.

PR #286, opened 2021-06-24
Release workflow: Install deps individually

tidoust:
To run the release script, the release workflow installed dependencies with an `npm ci` call. Problem is that this triggers the prepare script as well, applying CSS/IDL patches. If some patch is no longer valid because the underlying spec got modified in the meantime, the workflow fails as a result.

That's a pity because the workflow does not depend on patches in practice. This update installs deps that the workflow needs individually to avoid applying patches.

### DIFF
--- a/.github/workflows/release-package.yml
+++ b/.github/workflows/release-package.yml
@@ -27,7 +27,10 @@ jobs:
         ref: master
 
     - name: Install dependencies
-      run: npm ci
+      run: |
+        npm install --no-save @octokit/rest
+        npm install --no-save rimraf
+        npm install --no-save @jsdevtools/npm-publish
 
     - name: Release package if needed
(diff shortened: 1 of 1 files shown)


tidoust:
CI fails due to invalid patch, not related to the PR.

dontcallmedom (approved):
LGTM ; an alternative (if we're really worried about tracking new dependencies over time) would be to do something à la https://stackoverflow.com/a/59886568/139591, but feels like an overkill at this stage

tidoust:
That's a good idea. I'll keep that in mind for later. We could typically detect whether the code is running in a GitHub workflow using one of the [environment variables](https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables) that GitHub sets.

PR #291, opened 2021-07-01
Restore ComputedEffectTiming.currentIteration etc.

saschanaz:
This mostly reverts #285 to restore the now-missing member `currentIteration` and the inheritance from `EffectTiming`.

### DIFF
--- a/ed/idlpatches/web-animations-2.idl.patch
+++ b/ed/idlpatches/web-animations-2.idl.patch
@@ -1,31 +1,21 @@
-From c23c657bba54488f8223ef05634aa1a12bd172bb Mon Sep 17 00:00:00 2001
-From: Francois Daoust <fd@tidoust.net>
-Date: Thu, 24 Jun 2021 10:55:15 +0200
-Subject: [PATCH] Drop duplicate `fillMode` enum, fix `ComputedEffectTiming`
+From ea2e5beb27b21ea54c3633c0282e71f1d9e14393 Mon Sep 17 00:00:00 2001
+From: Kagami Sascha Rosylight <saschanaz@outlook.com>
+Date: Thu, 1 Jul 2021 13:35:40 +0200
+Subject: [PATCH] Drop duplicate `fillMode` enum definition
 
 The spec is a delta spec and re-defines the enum to change the meaning of one of
-its value. It also extends but also modifies the definition of the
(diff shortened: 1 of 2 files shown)


ed/idlpatches/web-animations-2.idl.patch
```diff
@@ -1,31 +1,21 @@
...
-     double playbackRate;
- };
- 
--partial dictionary ComputedEffectTiming {
-+dictionary ComputedEffectTiming {
-     CSSNumberish         startTime;
-     CSSNumberish         endTime;
-     CSSNumberish         activeDuration;
 @@ -38,8 +38,6 @@ partial dictionary ComputedEffectTiming {
      CSSNumberish?        progress;
```
dontcallmedom, line 20: we still need the other fields here?

dontcallmedom (changes requested):
`ComputedEffectTiming` still needs to be patched to take into account the overlap between the -1 definition and the -2 changes in the type of the fields.

dontcallmedom:
Once we converge on this, I think we should trigger a new IDL package release since the current one shouldn't be relied on

ed/idlpatches/web-animations-2.idl.patch
saschanaz: Not sure I follow? This stops patching ComputedEffectTiming in level 2 and instead do it in level 1.

dontcallmedom: I may easily be confused, since reviewing diff of patches is a bit hard to process for me; but my impression is that the two patches would end up with a consolidated `ComputerEffectTiming` dictionary without `startTime`, and with the wrong type on `endTime` and `activeDuration`

saschanaz: Since my patch still removes all fields from level 1 spec including `startTime`, `endTime` and `activeDuration` but not `currentIteration`, they should follow level 2 definition, which is consistent to the current situation (unless I'm confused by something).

dontcallmedom: ah, I see - thanks, I was indeed confused by the multiple mental patching.

dontcallmedom (approved):

tidoust (approved):

PR #297, opened 2021-07-07
Drop SVG 1.1 elements and IDL deprecated in SVG 2

### LINKED ISSUE #278: Elements extraction: unknown interfaces for some SVG 1.1 elements

tidoust:
See rationale in https://github.com/w3c/webref/issues/278#issuecomment-874168240

This update keeps SVG 1.1 elements that still exist and rolls back the IDL patch on SVG 2 to what it used to be before SVG 1.1 only interfaces got added to it.

Fixes #278.

### DIFF
--- a/ed/elementspatches/SVG11.json.patch
+++ b/ed/elementspatches/SVG11.json.patch
@@ -1,17 +1,19 @@
-From b439b3472862e5de769058366cc42f3480651d15 Mon Sep 17 00:00:00 2001
+From ffbc27e5547fc329522bb35a113d051d5a0ef7ec Mon Sep 17 00:00:00 2001
 From: Francois Daoust <fd@tidoust.net>
-Date: Fri, 18 Jun 2021 17:31:41 +0200
-Subject: [PATCH] Drop non implemented SVG 1.1 only elements
+Date: Wed, 7 Jul 2021 10:53:15 +0200
+Subject: [PATCH] Drop SVG 1.1 elements deprecated in SVG 2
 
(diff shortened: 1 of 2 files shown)


tidoust:
Note CI fails due to #296.

dontcallmedom (approved):

PR #300, opened 2021-07-10
Update workflows: adjust checkout branch for Reffy

tidoust:
Needed following switch to "main" in Reffy.

I note that most workflows currently reference live versions of actions, which isn't too good. I'll prepare another PR to address that.


tidoust:
@dontcallmedom I'll take the liberty to merge right away so that update actions can resume.

PR #317, opened 2021-08-04
Update portals patch to fix additional IDL inconsistency

dontcallmedom:
See WICG/portals#273 and WICG/portals#275 (patch has been merged but published spec not updated yet)

replace #315


dontcallmedom:
merging as hotfix to unblock updated IDL package generation

Issue #127, opened 2021-03-10
CSS package: consider additional parsing guarantees


tidoust:
See discussion in https://github.com/w3c/webref/pull/115#discussion_r590699701

tidoust:
Some notes as I'm looking into adding a check for unicity for CSS property definitions.

There are a number of properties that are defined more than once in practice. In most cases, this is "by design" and needs to be accounted for in the code:
- Many CSS 2.x properties are redefined in dedicated CSS modules. Definitions in modules take precedence.
- Delta specs are often not pure delta specs and may redefine a property.
- css-align > css-flexbox, see [Module interactions](https://drafts.csswg.org/css-align/#placement)

There remain properties defined in more than one spec, and for which I don't always know which one is the "right" one:
- `fill-opacity` defined in [fill-stroke-3](https://drafts.fxtf.org/fill-stroke/), [SVG2](https://svgwg.org/svg2-draft/)
- `fill-rule` defined in [fill-stroke-3](https://drafts.fxtf.org/fill-stroke/), [SVG2](https://svgwg.org/svg2-draft/)
- `fill` defined in [fill-stroke-3](https://drafts.fxtf.org/fill-stroke/), [SVG2](https://svgwg.org/svg2-draft/)
- `image-rendering` defined in [css-images-3](https://drafts.csswg.org/css-images-3/), [SVG2](https://svgwg.org/svg2-draft/)
- `inline-size` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [SVG2](https://svgwg.org/svg2-draft/)
- `inset-block-end` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [css-position-3](https://drafts.csswg.org/css-position/)
- `inset-block-start` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [css-position-3](https://drafts.csswg.org/css-position/)
- `inset-block` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [css-position-3](https://drafts.csswg.org/css-position/)
- `inset-inline-end` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [css-position-3](https://drafts.csswg.org/css-position/)
- `inset-inline-start` defined in [css-logical-1](https://drafts.csswg.org/css-logical-1/), [css-position-3](https://drafts.csswg.org/css-position/)
[...]

PR #327, opened 2021-08-19
Code to make RDF ontology from CSS properties JSON

timbl:


tidoust (commented):
Thanks. The code seems good but I'm not fully clear what value such an RDF ontology is going to bring if:

1. It only contains CSS properties from the CSS 2.2 specification (`../ed/css/CSS.json`), excluding CSS properties defined in other CSS module specs (the other JSON files in `../ed/css/`). What are usage scenarios that only need CSS 2.2 properties?
2. Its maintenance is not automated. In other words, https://www.w3.org/ns/css.ttl strikes me as a brittle source of truth if it gets updated manually. Automated maintenance is doable. I note we need to fix/improve a couple of things first (starting with making sure that CSS properties are defined only once, see https://github.com/w3c/webref/issues/127)

tidoust:
Going ahead and merging the PR. I filed #366 to track improvements and automation.
I'll add the editorial fixes after merge.

PR #343, opened 2021-09-07
Delete secure-contexts IDL extract

tidoust:
Spec was temporarily published with IDL that has already been integrated in HTML yesterday.

Plus the WebIDL is invalid (should be "partial interface **mixin**").


tidoust:
FWIW, IDL tests that fail are due to other IDL errors in other specs.

PR #346, opened 2021-09-07
Add patch for ed/idl/private-click-measurement.idl

tidoust:
Drop duplicate definition of attributionDestination.

Problem arises because browser-specs has both Private Click Measurement and Attribution Reporting, which seem to be "competing" proposals to achieve the same purpose and that both extend `<a>` elements with an `attributionDestination` attribute.

This just drops one. Perhaps we could drop one of the specs from browser-specs instead (but which one?).

### DIFF
--- a/ed/idlpatches/private-click-measurement.idl.patch
+++ b/ed/idlpatches/private-click-measurement.idl.patch
@@ -7,6 +7,8 @@ Private Click Measurement and Attribution Reporting are two "competing"
 proposals that both extend HTMLAnchorElement with an attributionDestination
 attribute. This drops one of the definitions while we get down to a situation
 where there is but one spec.
+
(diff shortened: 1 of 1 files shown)


dontcallmedom (commented):
LGTM, but what will be the trigger in removing that patch?

tidoust:
> LGTM, but what will be the trigger in removing that patch?

I don't necessarily have a good automatic trigger to suggest. I could create an issue on either or both repos, but if the final result is that one spec gets dropped, issues may never get closed in the end, and trigger won't work.

dontcallmedom:
how about raising an issue in browser-specs (since as you pointed out, this is a possible place we would "fix" it), and referring to it from the patch?

PR #352, opened 2021-09-10
Bump version of WebIDL2.js

tidoust:
Latest version validates `const` in `namespace`, which was allowed recently in Web IDL.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,1617 @@
 {
   "name": "webref",
   "version": "0.0.1",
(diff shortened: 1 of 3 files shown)


dontcallmedom (changes requested):
package-lock.json needs to be updated too

dontcallmedom (changes requested):
and the @webref/idl co-dependency to the WebIDL package needs to be updated too

dontcallmedom (approved):

PR #364, opened 2021-09-22
Get rid of "--delay" mocha CLI option

### LINKED ISSUE #363: Running `npm run test-idl` does not run tests if IDL cannot be parsed
See https://github.com/w3c/webref/issues/360#issuecomment-924876966_

In such cases, `parseAll` throws an error, and that error does not seem to be picked up by mocha.

There may be a conflict with the `--delay` CLI option. Actually, I'm not clear that our use of `run` is correct. The `run` function should be called once for the entire test suite, and not once per test file, I believe, and we're calling it multiple times (in CSS test files and in IDL test files). We may need to re-organize test files accordingly.

tidoust:
This is needed to fix #363. The `--delay` option requires the code to call the `run` function only once and we were calling it multiple times. More importantly, Mocha does not catch exceptions thrown at the root level and simply reports nothing when that happens. These exceptions typically get thrown when the IDL cannot be parsed, meaning that invalid data exceptions mostly went unnoticed...

This update wraps all code that needs to run before tests in a `before` function. Note the need to keep a "dummy" test at the root level, otherwise the `before` function won't run (and wouldn't have anywhere to report failure). This workaround is somewhat documented in mocha's repository, see https://github.com/mochajs/mocha/issues/2116#issuecomment-533211204

This update slightly improves describe/it tests naming as well.


tidoust:
Dammit, I inadvertently committed additional unicity tests for CSS that cannot pass today. Will rollback...

tidoust:
Tests now fail because they actually run, and detect that `startTime` is defined twice. We'll need to fix that through patches. Anyway, that's orthogonal to this PR.

test/idl/package.js
```diff
@@ -3,8 +3,8 @@ const assert = require('assert').strict;
 const idlPackage = require('../../packages/idl/package.json');
 const rootPackage = require('../../package.json');
 
-describe('@webidl/idl package', () => {
-  it('webidl2.js dependency version', () => {
+describe('The @webidl/idl package', () => {
```
dontcallmedom, line 6:
```suggestion
describe('The @webref/idl package', () => {
```

test/css/all.js
```diff
@@ -2,8 +2,8 @@ const assert = require('assert').strict;
 
 const css = require('@webref/css');
 
-describe('@webidl/css module', () => {
-  it('listAll', async () => {
+describe('The @webidl/css module', () => {
```
dontcallmedom, line 5:
```suggestion
describe('The @webref/css module', () => {
```

dontcallmedom (approved):
I'm not sure I fully understand the mocha magic, but I'm more than happy to trust that it works :)

A few nits worth fixing (even though they were not introduced by this PR)

PR #434, opened 2021-12-09
Fix name of reference for @latest branches

dontcallmedom:


dontcallmedom:
as listed in https://api.github.com/repos/w3c/webref/git/refs

tidoust (approved):

Issue #277, opened 2021-06-17
Release curated data on the Web


dontcallmedom:
#269 made it apparent that keeping NPM-only the consistency-guarantees provided by the NPM packages had two limitations:
* it forces to go through NPM which is at best awkward for non-NPM based consumers
* it limits the impact of the curation to whatever is packaged (e.g. the raw IDL files for the WebIDL package) - there is no easy way to access the generated JSON files that derive from it

An idea @tidoust and I discussed was to use w3c.github.io/webref to publish the curated-based view of the data.

This would require:
* moving the anomaly report out of webref (which we have discussed doing for quite some time in any case)
* adopting the release workflows to make publication on the `gh-pages` branch another of their outcomes

One question that needs more thinking is how to manage versioning.

tidoust:
Additionally, it would require generating the curated versions of folders in the first place. We don't produce the curated version of `idlnames` for the time being. It's just a matter of running the code on the curated version of the IDL, but that probably requires some adjustments to the code.

tidoust:
Following discussions with @dontcallmedom, updated plan is to only create one view in the `curated` branch that contains all specs, except that the `ed/idlnames` and `ed/idlnamesparsed` folders in that view will be restricted to actual browser specs. It should be trivial to filter out specs that people may not be interested in other folders.

Published npm packages will also continue to be restricted to actual browser specs.

tidoust:
I'm closing the issue as the curation job is now in place and seems to work. I wouldn't be too surprised if additional tweaks turn out to be needed but let's handle them separately.

The `@webref/idl@latest` ref will point to the right commit on the `curated` branch as soon as a new version of the `@webref/idl` package gets released (currently blocked by non-distinguishable IDL in Autoplay Policy Detection)

Same thing for the `@webref/css@latest` ref (there was a remaining bug in the release script when the `@webref/css@3.0.4` was released, so tags could not be added).

The `@webref/elements@latest` ref... does not exist yet, @dontcallmedom to create it and have it point to [`@webref/elements@1.0.4`](https://github.com/w3c/webref/releases/tag/%40webref%2Felements%401.0.4).

dontcallmedom:
`@webref/elements@latest` now created

PR #468, opened 2022-01-30
Update tools and jobs to publish curated data

tidoust:
The goal of this update is to create a curated data view along with the raw data view and the npm packages views (see #277).

Curation means applying patches to the raw data and re-generating the `idlparsed`, `idlnames` and `idlnamesparsed` folders. The latter two will only contain IDL names targeted at browsers, although note that actual spec filtering remains a TODO at this stage (see corresponding TODO comments in `prepare-curated.js` and `prepare-packages.js`).

To create the curated data view, this update introduces new tools:
- a `prepare-curated.js` tool that copies the raw data to the given folder, applies patches (CSS, elements, IDL) when needed, re-generates the `idlparsed` folder, re-generates the `idlnames` and `idlnamesparsed` folders and adjusts the `index.json` and `idlnames.json` files accordingly.
- a `prepare-packages.js` tool (replaces the now gone `packages/prepare.js`) that copies relevant curated data from the curated folder to the packages folder.
- a `commit-curated.js` tool that updates the `curated` branch with the contents of the given curated folder.

Goal is to have the `curated` branch be the one published as GitHub Pages.

The test logic was partially re-written to run the tests against the curated data, and against both the curated data and the NPM packages data when tests may yield different results.

A new `curate.yml` job publishes the curated data whenever the crawl data is updated. The job also takes care or preparing package release PRs as needed, replacing the previous prepare-xxx-release jobs.

The release workflow becomes:
1. Crawled data is updated (`update-ed.yml`)
2. Curated data and package data get generated (`curate.yml`)
3. Curated data and package data get tested (`curate.yml`)
4. The `curated` branch gets updated with the curated data (`curate.yml`)
5. Npm package pre-release PR gets created (`curate.yml`)
6. Someone reviews and merges the PR
7. New versions of npm packages are released (`release-package.yml`)
8. A `Raw data for @webref/ttt@vx.y.z` tag gets added to the relevant commit on the `main` branch.
9. A `@webref/ttt@vx.y.z` tag gets added to the relevant commit on the `curated` branch.
10. The `@webref/ttt@latest` tag gets updated to point to the relevant commit on the `curated` branch.

Note that, in order for a release to be created, curated data needs to have changed. A change to the static content in the `packages` folder won't be enough to trigger a release for instance. That should not be a major problem.


jodinathan:
this will help us so much :D

tidoust:
@dontcallmedom CI tests fail because the UUID spec still exists in webref. We should automate removal of files that are linked to a spec that got removed from browser-specs (in such cases, we can be confident that the removal was not accidental)

tidoust:
Replying to myself:

> Note that, in order for a release to be created, curated data needs to have changed. A change to the static content in the packages folder won't be enough to trigger a release for instance. That should not be a major problem.

Actually, that is going to be a problem because we want to do major/minor bumps under `packages` once in a while, and we want to update the package PR accordingly. I think that the easiest is to also include files under the `packages` folder in the `curated` folder so that updates to these files in the `main` branch also trigger updates to the `curated` branch. I'll update the scripts accordingly.

tools/prepare-curated.js
```diff
@@ -0,0 +1,146 @@
...
+  generateIdlNames, saveIdlNames,
+  generateIdlParsed, saveIdlParsed } = require('reffy');
+const {
+  createFolderIfNeeded,
+  requireFromWorkingDirectory,
+  copyFolder } = require('./utils');
+const { applyPatches } = require('./apply-patches.js');
+
+
+
```
dontcallmedom, line 29:
```suggestion
 /** 
*  Update crawl results to reflect successfully extracted data 
*/
```

tools/prepare-curated.js
```diff
@@ -0,0 +1,146 @@
...
+async function cleanCrawlOutcome(spec) {
+  await Promise.all(Object.keys(spec).map(async property => {
+    // Only consider properties that link to an extract
```
dontcallmedom, line 32: since the list of extracted properties is more or less hardcoded in the rest of the code (e.g. L77), it might be better to use a shared hardcoded list rather than rely on heuristics?

tools/utils.js
```diff
@@ -0,0 +1,80 @@
...
+ * @function
+ * @param {String} source The folder to copy
+ * @param {String} target The folder that should contain the copy. The folder
+ *    gets created if it does not exist.
+ * @param {Object} options Copy options. Set "excludeRoot" to exclude the root
+ *    folder from the copy and start the copy with its contents
+ * @return {Promise} The promise to have copied the folder's contents.
+ */
+async function copyFolder(source, target, options) {
+  options = options || {};
```
dontcallmedom, line 55: this too could use the deconstruction pattern to be more explicit about options

dontcallmedom (approved):
Yet another amazing piece of work, thank you so much!

A few nits for your consideration, but looks great to me in any case

packages/css/index.js
```diff
@@ -1,12 +1,14 @@
 const fs = require('fs').promises;
 const path = require('path');
 
-async function listAll() {
+async function listAll(options) {
+  options = options || {};
+  const folder = options.folder || __dirname;
```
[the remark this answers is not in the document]
tidoust: Done there and elsewhere. That syntax keeps bugging me as fairly unreadable but I suppose I'll get used to it at some point ;)

tools/prepare-curated.js
tidoust: I used slightly different wording, but done.

tools/prepare-curated.js
tidoust: I went the opposite way, actually, and dropped the hardcoded list (except to note the folders that must not be integrated), so that the script can handle new extracts without having to be modified.

Note the heuristics are exactly the same as in the `expandCrawlResults` function in Reffy.

tools/apply-patches.js
```diff
@@ -0,0 +1,111 @@
...
+    if (!type.includes(name)) {
+      continue;
+    }
+
+    await createFolderIfNeeded(dstDir);
+
+    // rm dstDir/*.${fileExt}
+    const dstFiles = await fs.readdir(dstDir);
+    for (const file of dstFiles) {
+      if (file.endsWith(`.${fileExt}`) && file !== 'package.json') {
```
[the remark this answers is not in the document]
tidoust: Tracked in #472.

PR #482, opened 2022-02-01
Do not include diff of index.js in pre-release PR

dontcallmedom:
Presumably we use the diff to focus on data review, not general package review

### DIFF
--- a/tools/prepare-release.js
+++ b/tools/prepare-release.js
@@ -70,7 +70,7 @@ function computeDiff(type) {
   // code to 0 to avoid the exception.
   const installedFiles = path.join(tmpFolder, "node_modules", "@webref", type);
   let diff = execSync(
-    `diff ${installedFiles} packages/${type} --ignore-trailing-space --exclude=package.json --unified=3 || echo -n`,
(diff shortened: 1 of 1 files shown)


tidoust:
I would argue that these files are in the diff by design?

If you exclude them from the diff, then no PR gets created when only these files change, even though a new release may be warranted. Typically, the `@webref/elements` package would not have received the JS updates if the diff had not included `index.js`.

`package.json` is excluded because there is no way to include it in the diff without triggering a diff due to the version bump, but ideally it would be considered as well.

dontcallmedom:
I wondered if this was by design; if so, I think still it would be best to separate the data diff from the rest, so maybe have different blocks of diffs - wdyt?

dontcallmedom:
(re `package.json`, we could distinguish the diff used to trigger the pull request from the diff that is displayed)

tidoust:
If it were me, I would lazily propose to just leave things as-is because these files only very seldom change, but feel free to do that!

dontcallmedom:
fine, I'll take the lazy approach; we'll lose the great improvement I had on L123 as an unfortunate consequence of that laziness

Issue #470, opened 2022-01-31
Automatically remove extracts from specs that no longer are in browser-specs


tidoust:
Via https://github.com/w3c/webref/pull/468#issuecomment-1025530895.

When a spec gets removed from browser-specs, we can be confident that the corresponding extracts can be safely removed from the repository. The update jobs should take care of these files immediately.

Note that once we have a mechanism in place to reuse last crawl data in Reffy (see https://github.com/w3c/reffy/pull/825), we should be able to fully automate the removal of files from Webref.

tidoust:
Semi-automation done. Full automation requires us to be truly confident that the fallback mechanisms really catches all corner cases where extracts cannot be produced.

PR #487, opened 2022-02-02
Use reffy fallback option to ignore erroneous data from errored crawls

dontcallmedom:


dontcallmedom:
did it only for the `ed` crawl, but would extend it to `tr` if/when it looks correct

tidoust (approved):
LGTM with nits.

On top of expanding to the `tr` crawl, once we're confident that results are good, we should also be able to replace the "Copy report" rsync command with a command that resets the target folder before it copies the files, so as to automatically remove files that are no longer needed (c.f. #470). The clean job could then be dropped altogether.

tidoust (changes requested):
Note the need to update package-lock.json as part of the PR otherwise `npm ci` fails

dontcallmedom:
can you update package-lock? when I npm install the latest reffy, it updates it to use lockfileVersion 1 which I assume we don't want

tidoust (approved):

PR #497, opened 2022-02-08
Remove non-browser specs from curated IDL names and npm packages

tidoust:
This update replaces the TODOs that had been envisioned with actual spec filtering to remove specs that don't target web browsers when creating the curated view of the `idlnames` and `idlnamesparsed` folder. These specs are also removed from the `@webref/xxx` npm packages.


tidoust:
CI tests fail because of #498. Not related to this PR.

dontcallmedom (approved):

PR #526, opened 2022-03-10
Make create-patch tool OS independent

tidoust:
The `mv` command does not usually exist in Windows environment.


tidoust:
Note CI tests fail for unrelated reasons.

dontcallmedom (approved):

PR #531, opened 2022-03-14
Use new API for octokit throttling plugin

dontcallmedom:
see https://github.com/octokit/plugin-throttling.js/commit/4c45c2e1eeb518e2a2c2f72d58526b755cee22ce


tidoust (approved):
Looks good. That should remove the warnings in the logs.
I note that this should not solve the quota exceeded errors that we seem to get once in a while though.

PR #551, opened 2022-04-07
[Tests] Add parsing exception for "cursor" CSS property

tidoust:
The `cursor` property definition now uses a stack combinator `#?`, which is not supported by the `css-tree` parser. Adding the property to the list of exceptions as a result.


dontcallmedom (commented):
did you check if the existing ones are still using non-parsed constructs?

It may be useful to add e.g. as a comment the construct that is known to be problematic to make it easier to review over time

tidoust:
I added some comments. It's either stacked combinators or the `,*` construct.

PR #555, opened 2022-04-09
Drop duplicate CSS property definitions when curating data

tidoust:
This adds a new data curation step that drops duplicate CSS property definitions from CSS extracts whenever possible, in other words when we know which definition is the authoritative one. The code contains a list of known superseding relationships between specs to choose the right one. See discussion in https://github.com/w3c/webref/issues/127 for details.

CSS properties that get re-defined in a delta spec are ignored, meaning that a naive merge of all curated CSS extracts will still contain such duplicates. A typical solution to end up with a consistent merge would be to exclude delta specs from that merge. This is left to data consumers as some may choose to rather live on the bleeding edge.

The curated data may still contain duplicate CSS property definitions, but these will get caught by tests. The new tests also check that CSS extracts contain a base CSS property definition for all CSS properties that get extended (through `newValues`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ Data curation brings the following guarantees.
 ### CSS extracts
 
 - All CSS files can be parsed by the version of [CSSTree](https://github.com/csstree/csstree) referenced in `package.json`, with the exception of a handful CSS value definitions that, although valid, are not yet supported by CSSTree.
+- No duplicate definitions of CSS properties provided that CSS extracts of [delta specs](https://github.com/w3c/browser-specs/#seriescomposition) are not taken into account (such extracts end with `-n.json`, where `n` is a level number).
+- CSS extracts contain a base definition of all CSS properties that get extended by other CSS property definitions (those for which `newValues` is set).
 
 ### Elements extracts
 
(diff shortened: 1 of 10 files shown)


dontcallmedom (approved):

tidoust:
Hmm, there is actually a bug in the code, because `require` is used to load JSON files throughout, and `require` caches objects by default. When `index.json` is loaded by the `drop-css-property-duplicates.js` module and CSS extracts are expanded, the next step that loads `index.json` actually sees the result of that expansion, and not the file as it exists on disk. I'll fix that.

Two additional notes:
- The only properties that remain in the `CSS.json` extract once this logic is in place are: `z-index`, `page-break-before`, `page-break-after`, and `page-break-inside`.
- The logic only covers properties, and does not look at valuespaces and descriptors. It would be good to look into that next.

tidoust:
I also completed the READMEs to mention the new guarantees.

The whole thing does not introduce any backward incompatibility, but should we still consider that this should trigger a major bump of the CSS package given that it drops (duplicate) property definitions from a number of CSS extracts?

dontcallmedom (approved):
+1 to major bump

PR #556, opened 2022-04-11
Drop most CSS value spaces duplicates

tidoust:
This applies the same logic as for CSS properties to drop duplicates from CSS extracts. Some duplicates remain: `<fit-content()>`, `<length>`, `<path()>`, `<percentage>`, `<rect()>` and `<url>`.

No easy way to get rid of them, as functions and terms seem to be reused on purpose with similar though slightly different meaning each time. These duplicates are hardcoded into tests.

No update to guarantees in README as it's not clear what interesting guarantees we can really make for now. Value spaces remain too messy to be directly usable, e.g. see https://github.com/w3c/webref/issues/127#issuecomment-1094811809

### DIFF
--- a/test/css/all.js
+++ b/test/css/all.js
@@ -43,11 +43,54 @@ const tempIgnore = [
   { shortname: 'svg-strokes', prop: 'valuespaces', name: '<dasharray>' }
 ];
 
+
+// Valuespaces that are defined more than once...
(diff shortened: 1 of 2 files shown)


tidoust:
Taking the liberty to merge to release a somewhat consistent CSS package.

Issue #445, opened 2021-12-21
Obsolete elements are not included in @webref/elements


foolip:
Elements listed in https://html.spec.whatwg.org/multipage/obsolete.html are not part of `@webref/elements`, and it would be nice if they were. Maybe with `"obselete": true` in there to allow filtering.

In particular, the fact that `<xmp>` uses `HTMLPreElement` would have been nice to be able to infer from this data.

tidoust:
I don't really see how we could automate the extraction of that information from the HTML spec but both the elements extract and this list are unlikely going to change much, so we could maintain an elements patch to add the info if that seems useful.

I'm not clear how to map obsolete elements to interfaces though. Do the alternatives suggested in the spec necessarily mean that the element uses the alternative interfaces? For instance:
- `applet`: does the element use `HTMLEmbedElement` or `HTMLObjectElement`? Or neither?
- `bgsound`: does the element really use `HTMLAudioElement`?
- `frame`, `frameset`, `noframes`: what interface does each of them use?
- `strike`: `HTMLModElement` (for `<del>`) or `HTMLElement` (for `<s>`)?

tidoust:
> I'm not clear how to map obsolete elements to interfaces though.

I'm still not clear whether the HTML spec mandates an official mapping. I tried to list the right interfaces, defaulting to `HTMLElement`, in the patch proposed in #569, using code such as the following to test whether my assumption was correct:

```js
// The HTML spec says xml must be treated as pre
// This returns true
document.createElement('xmp') instanceof HTMLPreElement

// The HTML spec says plaintext must be treated as pre
// ... but this returns false, so mapping to HTMLElement instead
document.createElement('plaintext') instanceof HTMLPreElement
```

foolip:
Can you try creating the elements by parsing instead of `document.createElement()`? There might be parser aliases confusing matters here.

dontcallmedom:
Here is what I get from parsing HTML obsolete elements:
APPLET HTMLUnknownElement
ACRONYM HTMLElement
BGSOUND HTMLUnknownElement
DIR HTMLDirectoryElement
NOFRAMES HTMLElement
ISINDEX HTMLUnknownElement
KEYGEN HTMLElement
LISTING HTMLPreElement
MENUITEM HTMLUnknownElement
NEXTID HTMLUnknownElement
NOEMBED HTMLElement
PARAM HTMLParamElement
RB HTMLElement
RTC HTMLElement
STRIKE HTMLElement
[...]

PR #569, opened 2022-04-22
Create patch to add obsolete elements in ed/elements/html.json

tidoust:
Fix for #445, also triggered by recent change in HTML spec that makes `param` an obsolete element, see https://github.com/w3c/webref/pull/566#issuecomment-1105998407

This adds obsolete elements to the HTML extract, in other words those defined in:
https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features

These obsolete elements cannot easily be extracted automatically but the list is unlikely to evolve significantly over time, and should only grow as new elements get obsoleted. When that happens, the element would first disappear from `elements/html.json`, which should provide the nudge we need to update this patch.

Obsolete element definitions come with an `"obsolete": true` property to distinguish them from regular HTML elements. Some of them have dedicated interfaces, and a couple of others inherit from specific interfaces (this is not very clear in the spec though, e.g. `xmp` implements the `HTMLPreElement` interface whereas `plaintext` does not, yet the prose in the spec is similar in both cases).

Note: Given the new `obsolete` property, we should do a major bump for `@webref/elements` when this gets released.

### DIFF
--- a/ed/elementspatches/html.json.patch
+++ b/ed/elementspatches/html.json.patch
@@ -1,0 +1,185 @@
+From cb31d9b6e0d8063d745687b268fe153ce5d21191 Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Fri, 22 Apr 2022 15:39:27 +0200
+Subject: [PATCH] Add obsolete HTML elements
+
+Patches the list of HTML elements with elements that are entirely obsolete but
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
LGTM although https://github.com/w3c/webref/blob/main/packages/elements/README.md probably needs an update (and +1 to major bump)

Issue #563, opened 2022-04-19
Handle CSS value spaces rgb() and hsl() with second "legacy" definition


tidoust:
[`<rgb()>`](https://drafts.csswg.org/css-color-4/#rgb-functions) and [`<hsl()>`](https://drafts.csswg.org/css-color-4/#the-hsl-notation) are defined with a legacy syntax that `@webref/css` does not extract. The legacy syntax defines a comma separated list of arguments while the modern syntax defines a whitespace/slash separated list. The legacy syntax will never be removed, I presume, and I would bet many CSS authors use the legacy syntax instead of the new one.

```html
<!-- Markup for the modern syntax -->
<pre class="prod highlight">
    <dfn class="dfn-paneled" data-dfn-type="function" data-export="" id="funcdef-hsl">
        <c- nf="">hsl</c-><c- p="">()</c->
    </dfn>
    <!-- ... -->
</pre>
<!-- Markup for the legacy syntax -->
<pre class="prod highlight">
    <c- nf="">hsl</c-><c- p="">()</c->
    <!-- ... -->
</pre>
```

I'm not sure what to suggest and where to suggest it. I believe that Bikeshed should allow defining a legacy syntax that `@webref/css` could assign to an additional `legacy` for an entry in `valuespace`, but to my knowledge, these are the only types that have two value definitions defined in the same spec.

_Originally posted by @cdoublev in https://github.com/w3c/webref/issues/306#issuecomment-1102438690_

cdoublev:
I have a preference for `legacyValue` because rewriting `modern` to `modern | legacy` would result to a different parse tree: `Or(Terminal('modern'), Terminal('legacy'))` vs `Terminal('modern')`. The 2nd solution could make a difference when the parse tree must be read from bottom to top, and a CSS grammar parser must have some sort of fallback mechanism anyway, eg. to handle math functions like `<calc()>` that can replace numeric CSS types.

PR #573, opened 2022-04-25
Add patch for ed/css/css-color.json

tidoust:
This adds legacy values for the `<rgb()>` and `<hsl()>` definitions.

The legacy values cannot easily be extracted from the prose (and if they could, we would have a hard time qualifying them as legacy definitions). Tools that want to handle legacy values need to join the `value` property and the `legacyValue` with a `|`.

Definitions are at:
https://drafts.csswg.org/css-color-4/#ref-for-legacy-color-syntax%E2%91%A1
https://drafts.csswg.org/css-color-4/#ref-for-legacy-color-syntax%E2%91%A3

This would fix https://github.com/w3c/webref/issues/563

Existential bump question: At a minimum, this should trigger a minor bump for `@webref/css`. Should this trigger a major bump? The new property adds new info but does not require changes in the way data consumers need to handle the data (unless they want to leverage the new info of course), so perhaps that's not needed?

### DIFF
--- a/ed/csspatches/css-color.json.patch
+++ b/ed/csspatches/css-color.json.patch
@@ -1,0 +1,44 @@
+From 7b815910e42fc0c102bc712a08808ec780b4aa59 Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Mon, 25 Apr 2022 11:50:02 +0200
+Subject: [PATCH] Add legacy values for rgb() and hsl() definitions
+
+The legacy values cannot easily be extracted from the prose (and if they were,
+we would have a hard time qualifying them as legacy definitions). Tools that
+want to handle legacy values need to join the `value` property and the
+`legacyValue` with a `|`.
+
+Definitions are at:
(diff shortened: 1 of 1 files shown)


dontcallmedom (commented):
so, this is a for-ever patch, right? We don't expect it to ever go away? I wonder if they should be managed differently; at least that probably should be made clear in the patch intro.

otherwise, +1 that a minor bump is sufficient

tidoust:
I clarified that the patch is meant to be a permanent one in the intro. Regarding managing this differently, I would prefer to continue using patches as long as that remains manage-able.

That's "only" the second one. First permanent patch was added last week for [HTML legacy elements](https://github.com/w3c/webref/blob/30e10c87daf73c27f93d3dd3183ca8af64715d80/ed/elementspatches/html.json.patch).

dontcallmedom (approved):
I wasn't too worried about using a permanent patch for HTML elements since they change so very rarely; the same probably won't hold true for css-color.json.

Let's proceed with this for now, and when maintaining the patch becomes too much of a pain, we can investigate alternative approaches :)

PR #577, opened 2022-04-27
CSS color patch: drop link to patch issue

tidoust:
The issue was for creating the patch, but the "Clean patches" job gets confused and assumes that it is the tracking issue for deleting the patch, see:
https://github.com/w3c/webref/pull/576


tidoust:
Note CI tests fail due to `AudioContext` duplicate hiccup, see https://github.com/WebAudio/web-audio-api/pull/2486

dontcallmedom (approved):

PR #584, opened 2022-05-02
Drop CSS property duplicates: add special case for css-logical

tidoust:
Via https://github.com/mdn/yari/pull/4656#issuecomment-1113508815

This ignores properties extended in css-logical: these extensions are against base definitions of CSS 2 properties, not against the new base definitions of these properties in more recent CSS modules.


dontcallmedom (commented):
wouldn't it be safer to list the 4 properties (`float`, `caption-side`, `clear`, `text-align` IIUC) instead of filtering out the spec completely?

dontcallmedom (approved):

Issue #630, opened 2022-06-16
Curate events data for publication as a package


dontcallmedom:
Following the addition to reffy in https://github.com/w3c/reffy/pull/952, webref now has a raw extract of events information as suggested in https://github.com/w3c/reffy/issues/772:
  https://github.com/w3c/webref/tree/main/ed/events

A goal should be to release that data through a new webref package (e.g. to help with https://github.com/openwebdocs/project/issues/61); before that we need to:
* agree on the invariants the package should fulfill
* massage the data to match these invariants

Based on discussion in https://github.com/w3c/reffy/issues/772#issuecomment-1122909270, I think we would want accurate data on event type, associated interface, interfaces that host the relevant event handler, and whether the event bubbles.

A strategy to achieve this would be:
* identify obvious gaps in the current data (usually available by detecting null values) and to the extent possible, fixing them upstreams in the relevant specs; for the remaining them, using our patch mechanism to patch the data in our processing pipeline
* ideally, review manually all the extracted data to check it matches what in the source spec (although we could also assume bad data will surface through consumer feedback if shipping early is preferable)

I'll follow up with details about the first of these bits - which specs provide only partial extractable data and what our options might be to make them extractable.

dontcallmedom:
Specs with partial data:
(skipping those for which there is already a pending PR)

# Missing information about the targets
 * [ ] [Content Security Policy Level 3](https://w3c.github.io/webappsec-csp/), `securitypolicyviolation` event
  * the event is described in CSP, but the targets are defined in HTML spec; possible solutions:
     * get the event targets described in CSP
     * identify some post-processing magic that maps the events across specs (since several events have a similar situation)
 * [ ] [CSS Spatial Navigation Level 1](https://drafts.csswg.org/css-nav-1/), "navbeforefocus" and "navnotarget" events
   * the spec [should define the related event handlers](https://github.com/w3c/csswg-drafts/issues/3380) (or have them added to the HTML spec, which would put us in a similar situation to CSP)
 * [ ] [CSS Scroll Snap Module Level 2](https://drafts.csswg.org/css-scroll-snap-2/), "snapChanged" and "snapChanging" events
    * as above
 * [ ] [EditContext API](https://w3c.github.io/edit-context/), "beforeinput" event
    * like CSP, target defined in HTML
 * [ ] [Import Maps](https://wicg.github.io/import-maps/), "error" and "load" events
[...]

PR #650, opened 2022-06-29
Adjust scripts to also create @webref/events package

tidoust:
This makes all needed adjustments to scripts for jobs to also release an `@webref/events` npm package.

Tests need to be adjusted to match guarantees that we can make. Guarantees will also need to be adjust in main README and in `packages/events/README.md`. See discussion in #630


tidoust:
@dontcallmedom I added tests for the guarantees we discussed (save the unicity of events with the same `type` and `interface` since I don't think we currently try to merge definitions). This triggers about 200 failures with the events extracts in the pull request (I suspect you may have fixed some of these failures through extraction improvements and pending PRs already)

dontcallmedom:
we're down to ~100 failures now, with already a few interesting issues resulting from the analysis: https://github.com/w3c/pointerevents/issues/451 https://github.com/w3c/gamepad/issues/166 https://github.com/WICG/speech-api/issues/103

(also discovered a new interface tree in the Serial API)

In terms of the proper way to model events:
* the fact that an event bubbles or not can be dependent on specific firing; more specifically, e.g. the `scroll` event can be firing bubbling on `Document`, and not bubbling on `VisualViewport` (the latter not being par of a tree means it can't bubble in any case); we could either relax the test on bubbling (e.g. is bubbles set when at least one of the interface is in a tree, with the (bold) assumption that bubbles would be ignored by the consumer when the interface is not in a tree), or we need to remodel the structure to associate bubbling with a particular target interface, e.g.:
```json
{ "type": "scroll",
   "targets": [ {"interface": "Document", "bubbles": true}, "VisualViewport" ]
}
```
  This could also / instead list the interfaces it can bubble to when bubbles is true?

* an invariant I thought we could test for - that a given event.type would always use a given event.interface on a given target does not hold, with [`success` being able to use either `Event` or `IDBVersionChangeEvent` on `IDBRequest`](https://www.w3.org/TR/IndexedDB/#ref-for-fire-a-success-event%E2%91%A0); I'm not sure if we should give up on this invariant altogether, or bring this to the TAG for discussion, possibly mangaging IDB as an exception in the end

tools/amend-event-data.js
```diff
@@ -22,6 +22,7 @@ const trees = {
   idb: ["IDBDatabase", "IDBTransaction", "IDBRequest"],
   bt: ['Bluetooth', 'BluetoothDevice', 'BluetoothRemoteGATTService',
        'BluetoothRemoteGATTCharacteristic', 'BluetoothRemoteGATTDescriptor'],
+  serial: ['Serial', 'SerialPort'],
```
tidoust, line 25: This makes me realize that the list needs to be maintained in two places (here and in tests), which isn't ideal.

Also, I'm not crazy about using a RegExp for element interfaces, since that means relying on a naming convention. Typically, I don't think `CSSPseudoElement` should be viewed as being part of a tree. Or should it? At least, it does not inherit from `Element` or `Node`.

Since the script already reuses the data in the `idlparsed` extracts, we should be able to follow the inheritance chain instead as done in the tests.

I'll update the script to remove the duplication and rather follow the inheritance chain.

tidoust:
That starts to look good. One thing that bugs me is that, currently, the curation process on events slightly changes the structure of the events extracts to convert `targets` from a list of strings to a list of objects with `target` and possibly `bubbles` properties, as well as to drop the `bubbles` property at the event level.

The curation process should only complete extracts, not change their structure. Curation should be something that can disappear in an ideal world where specs can be parsed automatically.

In other words, Reffy should rather generate the right data structure directly. Now, looking at Reffy's event extraction code, it seems clear that, from a spec perspective, whether the event bubbles and its list of target interfaces are orthogonal concepts. There will always be cases where Reffy can tell that an event bubbles but cannot determine target interfaces. The curated data structure does not really allow us to express that situation in a meaningful way.

Possible options on top of my head:
1. Good ol' times: Rollback on structural changes and get back to your idea of relaxing "the test on bubbling (e.g. is bubbles set when at least one of the interface is in a tree, with the (bold) assumption that bubbles would be ignored by the consumer when the interface is not in a tree)". No change to Reffy needed.
2. Cheese and dessert: Keep the `bubbles` property at the event level to mean that the event is, generally speaking, a bubbling event, but also report more fine-grained information at the `targets` level. Update Reffy to create the right `targets` structure.
3. No hand, no chocolate: Update Reffy to create the right `targets` structure and drop the `bubbles` info at the event level. We'll lose the bubbling info when Reffy cannot extract target interfaces, but that's up to curation to fix that.
4. Exception is the norm: Let the curation process change the structure as currently done, and avoid existential questions.

tidoust:
> can you say more on why you think we should ensure that curation doesn't change the structure? what does that commitment buys us?

You say that "manual curation is something we should be able to remove without impact". What this means to me is that the `main` and `curated` branches should remain as close to one another as possible, and converge towards being identical as quality of specs improves.

That's how we have structured curation so far. The `curated` branch just fixes data to provide additional guarantees but tools one may use to parse data in the `curated` branch can just equally run on the `main` branch. They just cannot assume that data will be valid or consistent. The `curated` branch and steps could disappear in a snap if patches are no longer needed. If we still need to run automatic conversion steps, we're creating a distinction between the `main` and `curated` branch that is not purely about data curation.

The way we've approached automatic conversion steps so far is that these steps happen during crawling and create separate extracts, and we re-run them during curation if needed. That's typically how we've approached `idlnames` and `idlparsed` for instance.

More importantly, I see the need to change the structure as a warning signal that we're doing something wrong. The change of structure seems motivated by the need to convey concepts that we cannot extract from specs today. There may be a time when the specs actually make these concepts explicit (and that seems a worthy goal). For instance, I'm wondering what is going to happen in your `scroll` example above when the monkey-patching in Visual Viewport will be upstreamed in CSSOM View. How are we going to convey that the `scroll` event bubbles on `Document`/`Element` but not on `VisualViewport` in the crawl results? We could perhaps create two entries for the event in the extract in that case, but then why don't we follow the same approach in the curated version?

Another more technical and less conceptual side argument towards keeping the same structure is that it eases the maintenance of event patches that we may need to create (in `eventpatches`, not the ones in `amend-event-data.js`). I would prefer to avoid having to mentally juggle between a before and after data structures when dealing with patches.

dontcallmedom:
> You say that "manual curation is something we should be able to remove without impact". What this means to me is that the main and curated branches should remain as close to one another as possible, and converge towards being identical as quality of specs improves.

The focus of my statement was on the manual bit (i.e. what is currently managed via `applyEventPatches`).

The `main` branch in my mind is something that we use only for convenience - a way to check what reffy extracted as a raw input to the preparation of the data we publish; the only constraint that should apply to that input is that it can be processed by the curation process. The ideal situation is where `main` is no longer needed, not when it matches `curated`.

> How are we going to convey that the scroll event bubbles on Document/Element but not on VisualViewport in the crawl results?

That's a fair question, but in practice:
* as far as I could determine, we don't have a reliable way of extracting that information per target interface in reffy
* we have a reliable way of detecting inconsistent information in the curation process

This to me illustrates the substantive bit of why I don't think we should constraint `main` and `curated` to use the same (or even simply "compatible") format: `curated` has by design access to more information which may require additional structure to express the additional granularity of information available compared to what can be extracted. If and when we can extract more information directly, I think re-using the format makes sense, but I wouldn't start from that basis when it doesn't gain us practical value.

(in this case, it also comes with a practical cost - e.g. it would make the DSL used to match events for patching more complex)

>  it eases the maintenance of event patches that we may need to create (in eventpatches, not the ones in amend-event-data.js)

I really don't think we should create patches using textual patches since we have the object-based patching mechanism that fulfils exactly that role. The event curation script purposefully applies these patches as the very first processing step to ensure the input to the patching process is clear (i.e. the content of the `main` branch).

Meanwhile, all the existing tests pass now :)

dontcallmedom:
> all the existing tests pass now 

Thinking aloud about possible additional invariants / tests:
* check that all targets (and their parents for bubbling events) expose the relevant event handler; maybe listing all the relevant event handlers in the data itself would be valuable
* check that the tuple (type, interface, target) is only defined once across all events

dontcallmedom:
another test (that would have caught a spec bug I only  discovered by accident): all the `*Event` interface are used by at least an event.

dontcallmedom:
another bit of potentially useful information (that would need associated tests): events whose handlers are also exposed as markup attributes.

(`WindowEventHandlers` have the particularity of [being exposed to content to despite being targeted at `Window`](https://html.spec.whatwg.org/multipage/webappapis.html#determining-the-target-of-an-event-handler) as I discovered in https://github.com/w3c/gamepad/issues/166)

tidoust:
> If and when we can extract more information directly, I think re-using the format makes sense, but I wouldn't start from that basis when it doesn't gain us practical value.

I don't understand that reasoning. Since we're the ones creating and maintaining both the crawl format and the curated format, and since we need to create the curated format to convey the information anyway, why not reuse the format right away for the crawl not to have to maintain two formats? This gains us practical value as in "one structure to maintain is better than two".

Even if we don't do that, I'd like us to make sure that the curated format can effectively be re-used. In practice, the curated format mixes the target interface and the bubbling information and we know that the crawl will have a hard time producing both info at once in the foreseeable future.

I'm thinking that a better format would actually be the one used by the crawler, using duplicate events in the list when needed. Data curation could then duplicate events that need to be, meaning those that both target interfaces that belong to a tree and those that don't. In that model, event identifiers would be the composed key `type`/`interface`/`bubbles`, instead of just `type`/`interface`. It's easy for consumers to consolidate duplicates if they would prefer to create a `type` or `type`/`interface` view, and consumers will likely need to do that to create consolidated views across specs in any case.

In other words, instead of:

```json
[
  {
    "type": "scroll",
    "interface": "Event",
    "targets": [
      { "target": "Element", "bubbles": true },
      { "target": "Document", "bubbles": true },
      { "target": "VisualViewport" }
    ]
  }
]
```

... we would end up with:

```json
[
  {
    "type": "scroll",
    "interface": "Event",
    "bubbles": true,
    "targets": ["Document", "Element"]
  },
  {
    "type": "scroll",
    "interface": "Event",
    "targets": ["VisualViewport"]
  }
]
```

... which is incidentally what the crawler already produces (in two extracts because the events are currently defined in two specs). If we ever feel the need to add the list of interfaces on which the event will bubble, I think we can still do that with that structure with an additional property at the event level (and if absolutely needed, we can duplicate the event again per target tree for events that can be fired at targets that belong to different trees).

I'm happy to update the curation script to produce the latter instead of the former.

Side note that, regardless of the above, we probably also should add another unicity guarantee along the lines of:
- check that the tuple (type, interface, bubbles (or not bubbles)) is only defined once per spec

dontcallmedom:
I'm sceptical that duplicating events across their bubbling status is the preferred outcome from a consumer perspective (e.g. I would expect MDN to want to document the "scroll" event uniquely, describing its bubbling behavior per interface). I guess that may be easier to assess with input from an actual customer :)

(semantically, I also think the bubbling status is not quintessential to the event, only to its handling by specific targets)

(The tests are failing again based on a new invariant check that is probably identifying bugs in the data)

tidoust:
> I'm sceptical that duplicating events across their bubbling status is the preferred outcome from a consumer perspective (e.g. I would expect MDN to want to document the "scroll" event uniquely, describing its bubbling behavior per interface). I guess that may be easier to assess with input from an actual customer :)

We can envision different scenarios:
1. Document events uniquely across specs. We're not providing a consolidated vision across specs for now, so consumers will have to consolidate events across specs regardless of the path we choose here.
2. Document events uniquely per spec. That would be what the current curated format provides.
3. Document events that fire on a specific target interface. To create such a view, a flat structure would be even better than one that merges targets together, meaning one with only a `target` property (and duplicated event entries for each target).

> (semantically, I also think the bubbling status is not quintessential to the event, only to its handling by specific targets)

I don't disagree although I note that various specs seem to report on the bubbling status in the definition of the event itself and in summary tables.

dontcallmedom:
Summarizing our discussion this morning (IIRC):
* reffy will now populate a `rawevents` (thinking about it - maybe `spec-events` instead?) folder that can be transformed into a single consolidated list of events via a reffy tool
* the patching system will remain in webref, and will be called prior to calling the consolidating / restructuring reffy library in the curation process - the end result is that the webref/events package should be a single JSON file listing all the known events
* the consolidation script won't commit its default reffy results on main@reffy; we will similarly remove idlparsed and idlnamesparsed from main@reffy
* the consolidated view will be per "event"

What I'm not sure we got a clear conclusion on:
* what constitutes identify for an event (proposed: type + interface?)
* how do we represent target interfaces and the fact the event bubbles on them or not (proposed: {target, bubbles})
* which target interfaces do we list (all the interfaces in the bubbling path, only the deepest) (proposed: only the deepest - mostly because that's what's currently implemented)

I think we said the best path forward was to land a first pre-release PR with a proposed approach, and uses it to seek input from others on the right way to structure the information.

dontcallmedom:
Another note from this morning: we need to ensure the way I use the bubbling-tree-info doesn't generate wrong data e.g. when targeted interfaces include non-HTML elements.

dontcallmedom:
> (The tests are failing again based on a new invariant check that is probably identifying bugs in the data)

These tests now pass, having allowed to identify several missing / mislabeled data (as well as a superfluously defined event interface)

dontcallmedom:
Exploring group events which share a type and an interface - here are the events that would end up being considered identical:
* abort using Event:
   - abort using Event in dom targeting AbortSignal
   - abort using Event in html targeting HTMLMediaElement
   - abort using Event in IndexedDB-3 targeting IDBTransaction
   - abort using Event in media-source-2 targeting SourceBuffer
   - abort using Event in uievents targeting Window,Element

* progress using Event:
   - progress using Event in html targeting HTMLMediaElement
   - progress using Event in background-fetch targeting BackgroundFetchRegistration

* error using Event:
   - error using Event in html targeting HTMLMediaElement
   - error using Event in html targeting HTMLSourceElement
   - error using Event in html targeting HTMLTrackElement
   - error using Event in html targeting Window,WorkerGlobalScope,EventSource,Worker,SharedWorker,Worker,SharedWorker,ServiceWorker
   - error using Event in notifications targeting Notification
   - error using Event in websockets targeting WebSocket
   - error using Event in IndexedDB-3 targeting IDBRequest
   - error using Event in media-source-2 targeting SourceBuffer
   - error using Event in uievents targeting Window,Element

* ended using Event:
   - ended using Event in html targeting HTMLMediaElement
   - ended using Event in mediacapture-streams targeting MediaStreamTrack
   - ended using Event in webaudio targeting AudioScheduledSourceNode

* pause using Event:
   - pause using Event in html targeting HTMLMediaElement
   - pause using Event in mediastream-recording targeting MediaRecorder

* resize using Event:
   - resize using Event in html targeting HTMLMediaElement
   - resize using Event in visual-viewport targeting VisualViewport
   - resize using Event in cssom-view-1 targeting Window
   - resize using Event in picture-in-picture targeting PictureInPictureWindow

* change using Event:
   - change using Event in html targeting AudioTrackList,TextTrackList,VideoTrackList
   - change using Event in html targeting HTMLInputElement,HTMLSelectElement
   - change using Event in window-placement targeting Screen
   - change using Event in window-placement targeting ScreenDetailed
   - change using Event in idle-detection targeting IdleDetector
   - change using Event in netinfo targeting NetworkInformation
   - change using Event in device-posture targeting DevicePosture
   - change using Event in permissions targeting PermissionStatus
   - change using Event in presentation-api targeting PresentationAvailability
   - change using Event in screen-orientation targeting ScreenOrientation

* load using Event:
   - load using Event in html targeting HTMLTrackElement
   - load using Event in html targeting Window
   - load using Event in uievents targeting Window,Document,Element

* cancel using Event:
   - cancel using Event in html targeting HTMLDialogElement,HTMLInputElement
   - cancel using Event in close-watcher targeting CloseWatcher

* close using Event:
   - close using Event in html targeting HTMLDialogElement
   - close using Event in notifications targeting Notification
   - close using Event in close-watcher targeting CloseWatcher
   - close using Event in IndexedDB-3 targeting IDBDatabase
   - close using Event in webrtc targeting RTCDataChannel

* message using MessageEvent:
   - message using MessageEvent in html targeting Window,EventSource,MessagePort,BroadcastChannel,DedicatedWorkerGlobalScope,Worker,ServiceWorkerContainer,PresentationConnection
   - message using MessageEvent in websockets targeting WebSocket
   - message using MessageEvent in portals targeting HTMLPortalElement
   - message using MessageEvent in portals targeting PortalHost
   - message using MessageEvent in webrtc targeting RTCDataChannel

* messageerror using MessageEvent:
   - messageerror using MessageEvent in html targeting Window,MessagePort,BroadcastChannel,DedicatedWorkerGlobalScope,Worker,ServiceWorkerContainer
   - messageerror using MessageEvent in portals targeting HTMLPortalElement
   - messageerror using MessageEvent in portals targeting PortalHost
   - messageerror using MessageEvent in service-workers-1 targeting ServiceWorkerGlobalScope

* open using Event:
   - open using Event in html targeting EventSource
   - open using Event in websockets targeting WebSocket
   - open using Event in webrtc targeting RTCDataChannel

* readystatechange using Event:
   - readystatechange using Event in html targeting Document
   - readystatechange using Event in xhr targeting XMLHttpRequest

* select using Event:
   - select using Event in html targeting HTMLInputElement,HTMLTextAreaElement
   - select using Event in uievents targeting Element

* unload using Event:
   - unload using Event in html targeting Window
   - unload using Event in uievents targeting Window,Document,Element

* icecandidateerror using RTCPeerConnectionIceErrorEvent:
   - icecandidateerror using RTCPeerConnectionIceErrorEvent in webrtc-ice targeting RTCIceTransport
   - icecandidateerror using RTCPeerConnectionIceErrorEvent in webrtc targeting RTCPeerConnection

* icecandidate using RTCPeerConnectionIceEvent:
   - icecandidate using RTCPeerConnectionIceEvent in webrtc-ice targeting RTCIceTransport
   - icecandidate using RTCPeerConnectionIceEvent in webrtc targeting RTCPeerConnection

* resume using Event:
   - resume using Event in page-lifecycle targeting Document
   - resume using Event in mediastream-recording targeting MediaRecorder

* connect using Event:
   - connect using Event in serial targeting Serial,SerialPort
   - connect using Event in presentation-api targeting PresentationConnection
   - connect using Event in remote-playback targeting RemotePlayback

* disconnect using Event:
   - disconnect using Event in serial targeting Serial,SerialPort
   - disconnect using Event in remote-playback targeting RemotePlayback

* start using Event:
   - start using Event in speech-api targeting SpeechRecognition
   - start using Event in mediastream-recording targeting MediaRecorder

* scroll using Event:
   - scroll using Event in visual-viewport targeting VisualViewport
   - scroll using Event in cssom-view-1 targeting Document,Element

* loadstart using ProgressEvent:
   - loadstart using ProgressEvent in FileAPI targeting FileReader
   - loadstart using ProgressEvent in xhr targeting XMLHttpRequest

* progress using ProgressEvent:
   - progress using ProgressEvent in FileAPI targeting FileReader
   - progress using ProgressEvent in xhr targeting XMLHttpRequest

* abort using ProgressEvent:
   - abort using ProgressEvent in FileAPI targeting FileReader
   - abort using ProgressEvent in xhr targeting XMLHttpRequest

* error using ProgressEvent:
   - error using ProgressEvent in FileAPI targeting FileReader
   - error using ProgressEvent in xhr targeting XMLHttpRequest

* load using ProgressEvent:
   - load using ProgressEvent in FileAPI targeting FileReader
   - load using ProgressEvent in xhr targeting XMLHttpRequest

* loadend using ProgressEvent:
   - loadend using ProgressEvent in FileAPI targeting FileReader
   - loadend using ProgressEvent in xhr targeting XMLHttpRequest

* devicechange using Event:
   - devicechange using Event in mediacapture-streams targeting MediaDevices
   - devicechange using Event in webxr targeting XRSystem

* statechange using Event:
   - statechange using Event in service-workers-1 targeting ServiceWorker
   - statechange using Event in webaudio targeting BaseAudioContext
   - statechange using Event in webrtc targeting RTCIceTransport
   - statechange using Event in webrtc targeting RTCDtlsTransport
   - statechange using Event in webrtc targeting RTCSctpTransport

* error using RTCErrorEvent:
   - error using RTCErrorEvent in webrtc targeting RTCDataChannel
   - error using RTCErrorEvent in webrtc targeting RTCDtlsTransport

There are a number of cases where merging these would seem sensible, but also quite a few where it's not clear that this would be meaningful (e.g. the semantics of "close" are very different between Notification and RTCDataChannel). I'm not sure what to make of it at this point yet.

tidoust:
> For instance, I'm wondering what is going to happen in your scroll example above when the monkey-patching in Visual Viewport will be upstreamed in CSSOM View.

Interestingly, the Visual Viewport API got folded into CSSOM View yesterday. Event extraction logic currently misses `VisualViewport` as a target because event dfn has not been properly updated (I filed a [PR to update the spec](https://github.com/w3c/csswg-drafts/pull/7479)).

Once spec is fixed, extraction will create one `scroll` event with a `bubbles` attribute set to `true` and targets `Document`, `Element`, `VisualViewport` (Curation will drop `bubbles` for `VisualViewport` since it is not part of a tree hierarchy). Given the `fire an event` phrasing used in the spec, the extraction could easily detect that the event is fired in two contexts, one where it bubbles, and one where it does not. It won't be easy to associate these contexts with relevant target interfaces though.

dontcallmedom:
> There are a number of cases where merging these would seem sensible, but also quite a few where it's not clear that this would be meaningful (e.g. the semantics of "close" are very different between Notification and RTCDataChannel). I'm not sure what to make of it at this point yet.

So ideally, the merging would be done on the basis of what specs expose (. The problem is that the way events are namespaced with `dfn-for` ties it to target interfaces, which is at odd with our proposed identity scheme (where target interfaces don't play a role). In fact, this makes it actually harder or awkard to re-use an event in a new context (as illustrated by the `VisualViewport` example above). Changing that convention feels like a challenging uphill battle though.

Looking at the list above, a reasonable first step would be to group events that share a name and a non-default interface (i.e. all events that don't use `Event`) - that is a strong enough signal that they're really about the same kind of events.

But I don't think we can easily go beyond that without deeper rethinking of how events are surfaced in specifications, which is probably out of scope for a 1.0 of the package.

dontcallmedom:
> Tests that still fail are due to missing target interfaces. More patches needed? Better consolidation algorithm? Or did I break something along the way?

Something definitely broke along the way; e.g. the errors include lack of targets for `canmakepayment` in `payment-handler`:
>  contains valid target interfaces for events "canmakepayment":
>     AssertionError [ERR_ASSERTION]: No target interfaces for event type=canmakepayment interface=CanMakePaymentEvent href=https://w3c.github.io/payment-handler/#the-canmakepaymentevent

while the [raw event has that info](https://github.com/w3c/webref/blob/0f6e400b0d007a7426e34339c887f5f5f100c1d2/ed/events/payment-handler.json#L14).

dontcallmedom:
should we merge and use the 1.0.0 package pull request to identify potential additional improvements?

tidoust:
I think we said that, for a bubbling event, `targets` should only contain the deepest target interface in a tree. I just added a test and that does not seem to be the case today. Consolidation does not seem to do what it should, I'll look into it.

dontcallmedom:
> Am I right that it should rather keep all listed target interfaces here because they are more specific than the core ones in the tree?

I think so, although I'm not sure that's a rule that will hold systematically either

PR #674, opened 2022-07-18
Adjust jobs for Reffy v8.0.0

tidoust:
This adjusts crawl, curation and clean jobs to account for post-processing modules introduced in Reffy v8.0.0.

The `csscomplete` post-processing module is run during the crawl because we created CSS patches and de-duplication logic with completed CSS extracts in mind. We could revisit that later on but that does not seem to be a priority.

The `idlparsed` and `idlnames` post-processing modules get run after IDL patches have been applied to generate the `idlparsed`, `idlnames`, and `idlnamesparsed` folders. These folders are no longer generated during the crawl, and thus no longer need to be monitored for things to clean (we'll still need to actually remove the folders from Git, I'll do it when first crawl with v8.0.0 of Reffy will have completed).

This update does not run the `events` post-processing module. To be included in #650 along with the patch mechanism for events.

The curation logic leverages Reffy's new ability to copy crawl results to another output folder. The whole logic could perhaps be further streamlined to call Reffy's `crawlSpecs` function only once, e.g. by running patch mechanisms as post-processing modules instead of treating them separately. That also does not seem to be a priority...


tidoust:
Tests fail on line endings in IDL extracts. I need to look into it. One day, I'll get that right :(

dontcallmedom (approved):

PR #682, opened 2022-07-23
Release PR: skip diff when npm package cannot be installed

tidoust:
The `@webref/events` package does not yet exist on NPM, and creating a release pull request for it would not work because first release needs to be manual. This update makes the release PR script skip the package when it cannot install a package.


tidoust:
Taking the liberty to merge to fix curation failures.

PR #713, opened 2022-08-25
Add patch for ed/css/mathml-core.json

dontcallmedom:
Patch CSS Extract from MathML Core


dontcallmedom:
(the remaining CI are errors are unrelated and should be fixed by #712)

PR #725, opened 2022-09-06
Upgrade to reffy v10

dontcallmedom:


tidoust (approved):
LGTM.

Breaking change documented in:
https://github.com/w3c/reffy/commit/bedcd23a98ef4ea384decb7e83bb63e557c0afe2

In particular:
"The breaking change only affects code that makes use of links extracts. These extracts are now using an object with an anchors property that lists the said anchors, instead of being listed directly as an array."

Issue #704, opened 2022-08-19
Improve changelog (at least for major bumps)


dontcallmedom:
we had two customers seeking clarity on the major bump of our latest @webref/css, which I take as a signal we should be more explicit about these :)
https://github.com/w3c/webref/pull/688#issuecomment-1220342208
https://github.com/mdn/yari/pull/6913#issuecomment-1220734477

tidoust:
Right, an artefact of our release workflow is that the [commit that bumps the version](https://github.com/w3c/webref/commit/2a42be76fcd2c138f96ea4a3b9456be0f880570d) and that highlights the breaking change isn't directly visible when the package gets released.

Now, to improve the changelog, we would need to have a changelog in the first place ;)

tidoust:
That's not perfect but we do have a [changelog](https://github.com/w3c/webref/blob/main/packages/css/CHANGELOG.md) now that documents major version bumps in general, and breaking changes in particular.

Issue #657, opened 2022-07-04
Validate extracts against schemas


tidoust:
With patch mechanisms being used on a recurring basis, it becomes easier to produce extracts that do not respect the initial extract's schema. It would also be good to document that schema since the schemas are only roughly described in package READMEs for now.

PR #731, opened 2022-09-14
Add schemas for all JSON extracts

tidoust:
This provides a first level of schema validation for curated data extracts, see #657 for context.

Goal is to make it easier to detect and document (through a changelog, so also useful for #704) situations where we change the structure of data extracts.

Schemas, notably those that deal with parsed IDL structures, could go deeper into details.

Tests are run against the curated version of data. That is not necessary for extracts that aren't actually curated (dfns, headings, ids, links, refs), just more convenient not to have branching logic in the test code.

Creating the PR as pull request as 69 of the new tests currently fail, either because extraction logic in Reffy needs to be slightly improved to create more consistent data structures, or because of actual issues in the specs themselves (e.g. invalid URL fragments).


tidoust:
Superseded by #749.

PR #737, opened 2022-09-19
Add patch for ed/idl/web-animations.idl

tidoust:
I completely missed the fact that there already existed an IDL patch for Web Animations. Previous PR #736, which was merged, overwrote that patch by mistake. The previous updates are still needed. That explains why CI tests failed, and should have been yet another signal that something was wrong ;)

This new PR merges the previous patch and the new one.


tidoust:
Taking the liberty to merge. CI tests now pass, and it cannot make things worse than before in any case ;)

PR #744, opened 2022-09-22
Adjust event patches for the "slotchange" event

tidoust:
Goal is to make sure that the `slotchange` event is listed as firing at a `HTMLSlotElement` element and bubbles, instead of firing at `ShadowRoot`.

This will lead to the following event definition in the consolidated list:

```json
{
  "href": "https://html.spec.whatwg.org/multipage/indices.html#event-slotchange",
  "src": {
    "format": "summary table",
    "href": "https://html.spec.whatwg.org/multipage/indices.html#event-slotchange"
  },
  "type": "slotchange",
  "targets": [
    {
      "target": "HTMLSlotElement",
      "bubbles": true,
      "bubblingPath": [
        "Node",
        "Document",
        "Window"
      ]
    }
  ],
  "interface": "Event",
  "extendedIn": [
    {
      "spec": "dom",
      "href": "https://dom.spec.whatwg.org/#ref-for-concept-event-fire%E2%91%A5"
    }
  ]
}
```

Treatment of `ShadowRoot` could still be improved: the bubbling path does mention `Node`, so event can indeed fire at `ShadowRoot`. That said, the `slotchange` is the only event that can fire at `ShadowRoot`, which seems worth calling out. I don't know how to do that with the current structure and logic though.

### DIFF
--- a/tools/amend-event-data.js
+++ b/tools/amend-event-data.js
@@ -161,6 +161,16 @@ const patches = {
       delete: true
     }
   ],
+  'dom': [
+    // Forget about `ShadowRoot` target because de-deplucation logic cannot tell
+    // that `HTMLSlotElement` is deeper than `ShadowRoot`, see discussion in
+    // https://github.com/w3c/webref/pull/729#issuecomment-1254048215
+    {
+      pattern: { type: 'slotchange' },
+      matched: 1,
+      change: { targets: null }
+    }
+  ],
   // pending https://github.com/w3c/edit-context/pull/31
   'edit-context': [
     {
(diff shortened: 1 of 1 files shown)


tools/amend-event-data.js
```diff
@@ -543,7 +561,12 @@ function applyEventPatches(spec) {
           continue;
         } else if (patch.change) {
           for (const target of Object.keys(patch.change)) {
-            event[target] = patch.change[target];
+            if (patch.change[target]) {
```
dontcallmedom, line 564:
```suggestion
            if (patch.change[target] !== null) {
```
Shouldn't this check specifically for `null` equality? otherwise this would have an expected result e.g. on `change: {bubbles: false}`

PR #760, opened 2022-10-11
Add patch for ed/idl/webaudio.idl

tidoust:
Rollback IDL pending resolution of https://github.com/WebAudio/web-audio-api/issues/2514

PR #761, opened 2022-10-11
Adjust event patches (incl. DOM and HTML)

tidoust:
This adjusts the list of event patches to account for the latest revisions of DOM, HTML, Screen Orientation and Web Audio API. This also drops the patch for import-maps, now folded in HTML spec, and adds a patch for screen-capture that explicitly does *not* fire the `devicechange` event it mentions.

### DIFF
--- a/tools/amend-event-data.js
+++ b/tools/amend-event-data.js
@@ -161,16 +161,6 @@ const patches = {
       delete: true
     }
   ],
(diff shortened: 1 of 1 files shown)


tidoust:
CI tests won't pass until this PR and PR #760 are merged (and until next crawl of the File System spec)

PR #797, opened 2022-11-24
Adjust code for upcoming new CSS extracts structure

tidoust:


dontcallmedom (approved):
shall this be marked as draft to avoid inadvertent merging until reffy 11 is released?

tidoust:
I bumped the version of Reffy, re-generated the CSS patches to account for the new structure, added a new CSS patch for the CSS 2.2 spec and updated the CSS extract of the css-pseudo spec to match the extract that will be produced with the latest version of Reffy.

All tests green now and this should unblock data curation.

PR #823, opened 2022-12-15
Relax test on elements extracts

tidoust:
The `interface` property may not always be set, even if one only considers the first item in the list.


tidoust:
Note CI tests will fail (unrelated, due to a duplicated IDL term)

PR #839, opened 2023-01-09
Filter out specs not in good standing from packages

tidoust:
Follows from the recent addition of the `standing` property in browser-specs. Specs not in good standing should be filtered out from packages.


dontcallmedom (approved):
LGTM - I'm guessing the CI error is because it's not using the latest browser-specs?

tidoust:
> LGTM - I'm guessing the CI error is because it's not using the latest browser-specs?

Yes, since crawl results don't yet have the `standing` property, this update basically filters out all specs from the packages, and tests fail because they're not expecting packages to be empty. This should work again once new crawl results are available.

PR #866, opened 2023-01-30
Drop IDL patch for Orientation Event

tidoust:
Spec fixed upstream. The `PermissionState` enum has been finally removed.


tidoust:
Note CI tests fail because of respec.org current hiccup.

PR #917, opened 2023-03-15
Drop CSS patch for CSS Backgrounds 4

tidoust:
Definition of property changed in the spec.


tidoust:
Not enough to make all tests pass, but that's unrelated.

PR #926, opened 2023-03-23
Sync cssom-view.idl patch with upstream change

dontcallmedom:


dontcallmedom:
merging, patch now applies (CI breaks for another patch)

PR #936, opened 2023-03-31
Use `::backdrop` definition from CSS Position 4

tidoust:
Core definition of the `::backdrop` pseudo-element is now to be found in the CSS Position 4 spec. Fullscreen merely extends the definition.


tidoust:
> are these duplicates flagged in the fullscreen repo?

Now they are, see https://github.com/whatwg/fullscreen/issues/221

PR #954, opened 2023-04-27
Adjust reference to npm-publish

tidoust:
Not advertized in the list of breaking changes made in v2 AFAICT, but main function is no longer exported as default.

Note: Latest version of NPM packages could not be published as a result, I'll try to do that manually.

### DIFF
--- a/tools/release-package.js
+++ b/tools/release-package.js
@@ -9,7 +9,7 @@ const path = require("path");
 const os = require("os");
 const { execSync } = require("child_process");
 const { rimraf } = require("rimraf");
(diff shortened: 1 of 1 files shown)


tidoust:
> Note: Latest version of NPM packages could not be published as a result, I'll try to do that manually.

I rather used the usual automated workflow because release script gets confused on my Windows machine with line endings when it applies patches (I'll look into that one day). This means that we skipped a patch version, jumping from `3.33.1` to `3.33.3` for `@webref/idl` for example. `3.33.3` is a much better version number than `3.33.2` in any case, isn't it?

PR #977, opened 2023-05-24
Delegate Web IDL consistency tests to Strudy

tidoust:
(@foolip FYI since you authored the initial Web IDL tests)

With this update, all Web IDL consistency tests are now run directly by Strudy. This allows to keep the test code simple and makes it possible to reuse the test logic in other contexts.

Nothing was lost in the switch. Guarantees still hold in particular. Strudy checks all IDL features that used to be checked in `consistency.js`, reporting anomalies in a more digestable form (which should make it slightly easier to find the root of a problem when one is found). Corresponding code in Strudy may be found in https://github.com/w3c/strudy/blob/main/src/lib/study-webidl.js

Strudy reports a couple of Web IDL issues that the previous code did not detect, namely the fact that the `Screen` interface needs to inherit from `EventTarget`, for use in the Window Management API, and the fact that the `*Track` interfaces need to be exposed on `DedicatedWorker`, for use in Media Source Extensions. These issues have been acknowledged by the groups that develop the specifications but the patch needs to be done against base interfaces that are under the control of other groups, and it's not clear to me that these other groups have acknowledged the issues yet. I did not feel like patching the IDL as a result. The tests currently explicitly ignore these issues for now in the interest of not blocking new package releases because of this.

Note: CI tests will currently fail because of the missing `EventTarget` inheritance on `SFrameTransform` (which was also not previously detected). I prepared https://github.com/w3c/webrtc-encoded-transform/pull/179. If that does not get merged soon-ish, I'll create a patch for this one on our side.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "mocha": "10.2.0",
         "reffy": "13.0.0",
         "rimraf": "5.0.1",
+        "strudy": "^2.0.0",
         "webidl2": "24.3.0"
       },
       "engines": {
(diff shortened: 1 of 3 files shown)


dontcallmedom (commented):
should we land this without the EventTarget inheritance check? This is after all not among the guarantees we commit to for the package…

tidoust:
> should we land this without the EventTarget inheritance check? This is after all not among the guarantees we commit to for the package…

I tried to resist the temptation but since I don't how we can enforce this across specs due to MSE and window-management, I added both the check on EventTarget on incompatible exposure sets to the ignorable list. My assumption is that these problems will be reported in a near future through Strudy directly.

dontcallmedom (approved):

foolip (commented):
Cool!

PR #1040, opened 2023-09-15
Add patch for ed/css/css-gcpm.json

tidoust:
Rolling back definition of `:nth()` pending support in CSSTree for `<an+b>` or some other resolution.

(CI tests will continue to fail because of an edit-context event issue. That's unrelated)


dontcallmedom (approved):
it's not technically a rollback, since it is now identifying a selector (when the previous version had it mistakenly as a function); but I don't think it matters?

tidoust:
I noted that it's a partial rollback in the patch. I don't think it matters indeed.

PR #1041, opened 2023-09-19
Adapt patches on clipboard events to changes in the spec

dontcallmedom:


dontcallmedom:
(the CI still fails for other independent remaining errors)

tidoust (approved):

PR #1056, opened 2023-10-12
Add patch to avoid overladed IDL definition in CSS View Transitions

dontcallmedom:


dontcallmedom:
the CI error is for a separate issue (addressed in #1057)

tidoust (approved):

PR #1062, opened 2023-10-19
Update patch to avoid overladed IDL definition in CSS View Transitions

dontcallmedom:


tidoust (approved):

tidoust:
I note Level 2 now needs fixing as well...

PR #1076, opened 2023-11-08
Add patch for ed/idl/webxr-plane-detection.idl

tidoust:
Add missing `[Exposed]` extended attribute


tidoust:
Note: CI tests fail for unrelated reason, addressed in #1077.

dontcallmedom (approved):

PR #1077, opened 2023-11-08
Adjust event patch for `pagereveal` in css-view-transitions-2

tidoust:
Interface name was changed in the spec, and firing wording improved. A patch is still needed though.


tidoust:
Note: CI tests fail for unrelated reasons, addressed in #1076.

dontcallmedom (approved):

PR #1084, opened 2023-11-14
Add event patch for orientation-event spec

tidoust:
The spec proxies "fire an event" through a custom "fire an orientation event" algorithm. This hides it from Reffy.

### DIFF
--- a/tools/amend-event-data.js
+++ b/tools/amend-event-data.js
@@ -257,6 +257,15 @@ const patches = {
       change: { interface: "NotificationEvent" }
     }
   ],
+  'orientation-event': [
+    // Spec uses an indirect "fire an orientation event" algo to fire events:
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
I would have expected we would have detected there was no event matching the interface `DeviceOrientationEvent` - do we only have the reverse test?

tidoust:
> I would have expected we would have detected there was no event matching the interface `DeviceOrientationEvent` - do we only have the reverse test?

We have both tests, and one of the failed tests was due to there being no event matching the interface `DeviceOrientationEvent`:
https://github.com/w3c/webref/actions/runs/6860079808/job/18653320897#step:6:12209

dontcallmedom:
Sorry for the noise, I hadn't understood this was addressing a recent change and failure

PR #1113, opened 2023-12-21
Regenerate patch for ed/idl/webgl1.idl

tidoust:
The IDL of WebGL1 was slightly updated yesterday, breaking the patch. The patch is still needed though.


tidoust:
Note: Updated patch works correctly, CI tests now fail after curation due to another issue in another spec.

PR #1121, opened 2024-01-08
Transferable `RTCDataChannel`

turansky:
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel)

### DIFF
--- a/ed/idl/webrtc.idl
+++ b/ed/idl/webrtc.idl
@@ -503,7 +503,7 @@ enum RTCSctpTransportState {
   "closed"
 };
 
-[Exposed=Window]
+[Exposed=Window, Transferable]
 interface RTCDataChannel : EventTarget {
   readonly attribute USVString label;
   readonly attribute boolean ordered;


tidoust:
Thanks @turansky! I understand why you might want to add the `[Transferable]` IDL extension attribute. This is not the right place to do so though. The `el/idl/webrtc.idl` extract contains the IDL definitions of the WebRTC spec, and that spec **does not** define [`RTCDataChannel`](https://w3c.github.io/webrtc-pc/#webidl-1143016005) as transferable. The extract does not need fixing.

The spec itself might. [Transferable data channels](https://w3c.github.io/webrtc-extensions/#rtcdatachannel-transferable) are currently defined in the WebRTC Extensions spec. That spec is a mixed bag of proposals with different and sometimes unclear levels of support within the WebRTC Working Group. We haven't added the WebRTC Extensions spec to the list of specs crawled to populate Webref for that reason. At least not yet, see discussion in https://github.com/w3c/browser-specs/pull/1119.

Typically, MDN chose to document the transferable nature of data channels but I don't know whether there is consensus within the WebRTC Working Group about this particular feature. @dontcallmedom might know?

tidoust:
Just realized that the [browser-specs PR](https://github.com/w3c/browser-specs/pull/1119) mentioned above is about [Media Capture and Streams Extensions](https://w3c.github.io/mediacapture-extensions/), not [WebRTC Extensions](https://w3c.github.io/webrtc-extensions/). Sorry for the confusion. Anyway, same thing for both specs: they haven't been added to browser-specs (and to Webref) yet because of their unofficial nature.

Issue #642, opened 2022-06-24
Include links in idlparsed, css, etc.


gsnedders:
Currently you can derive links by processing the data in `dfns`, however this isn't possible if you're trying to only use npm-released data (especially in the css case—I'm aware there's no release of idlparsed).

For example, for a property, it would be nice if the URL to the definition was included within the data in `css`.

meyerweb:
@gsnedders, are you suggesting something like the following, where the ID is stored in the object?  Or the whole URL?

```
{
  "spec": {
    "title": "CSS Table Module Level 3",
    "url": "https://drafts.csswg.org/css-tables-3/"
  },
  "properties": [
    {
      "id": "propdef-border-collapse",
      "name": "table-layout",
      "value": "auto | fixed",
      "initial": "auto",
      "appliesTo": "table grid boxes",
      "inherited": "no",
      "percentages": "n/a",
      "computedValue": "specified keyword",
      "canonicalOrder": "per grammar",
      "animationType": "discrete",
      "styleDeclaration": [
        "table-layout",
        "tableLayout"
      ]
    },
    {
…
```

gsnedders:
> @gsnedders, are you suggesting something like the following, where the ID is stored in each property’s object? Or did you mean the whole URL?

Probably the whole URL, because we probably don't want everyone having to implement the logic of "how to construct a URL for this" (especially in the face of multipage specs).

tidoust:
Thanks for the nudge, that proved useful to push the issue to the top of the pile ;)

CSS extracts in Webref now include an `href` key for all entries (except ones that extend a base definition defined in another spec). Also deployed in the `@webref/css` package, starting with [v6.11.0](https://www.npmjs.com/package/@webref/css/v/6.11.0). Let me know if you spot any issue!

I'll look into doing the same thing for idlparsed extracts.

tidoust:
All extracts that didn't have more specific links back to the spec now have such links in `href` properties attached to most objects: CSS extracts, idlparsed extracts, and elements extracts. Other extracts had these links already (or cannot have them in the case of the idl extracts).

For idlparsed extracts, some entries don't have links, usually because the spec does not have the appropriate definitions, see https://github.com/w3c/reffy/pull/1493 for details.

Closing the issue as addressed.

PR #1140, opened 2024-01-31
Add new guarantee for CSS package on links back to the spec

tidoust:
Linked to #642.


tidoust:
One side question is whether the update warrants a major release of the npm package. A minor release may be enough: the schema of the CSS extract changes, but it's not a breaking change per se (just a new `href` key, no structural change).

dontcallmedom (approved):
and +1 to just a minor bump

PR #1209, opened 2024-04-09
Patch out __TO_BE_MERGED from sanitizer-api.idl

foolip:
How to remove them from the spec is not entirly clear:
https://github.com/WICG/sanitizer-api/issues/216


tidoust:
Patch should no longer be needed as the spec now uses `class="extract"` to exclude the IDL fragments from the crawl.

Issue #1212, opened 2024-04-16
cancel event doesn't bubble


dontcallmedom:
see https://github.com/orgs/mdn/discussions/665#discussioncomment-9125774

tidoust:
But the spec says it does, at least for input elements?
https://html.spec.whatwg.org/multipage/input.html#common-input-element-apis:event-cancel

dontcallmedom:
but not on dialog elements? https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element:event-cancel

if so, not sure how we can represent this usefully…

dontcallmedom:
hmm... I misread, it's on the CloseWatcher interface that it doesn't bubble (it couldn't anyway)

dontcallmedom:
now there is a question as to why we don't detect the cancel event on CloseWatcher

dontcallmedom:
sorry, I'm confusing two data points:
* https://html.spec.whatwg.org/multipage/interactive-elements.html#the-dialog-element:event-canceI says it doesn't bubble on dialog
* https://html.spec.whatwg.org/multipage/interaction.html#the-closewatcher-interface:event-cancel-4 says that it doesn't bubble on CloseWatcher (which we're not tracking at all)

tidoust:
> now there is a question as to why we don't detect the cancel event on CloseWatcher

That's because the `data-dfn-for` attribute needs fixing in the [definition of `cancel`](https://html.spec.whatwg.org/multipage/indices.html#event-cancel) in the HTML spec. It should be `data-dfn-for="CloseWatcher,HTMLElement"` and not only `data-dfn-for="HTMLElement"`.

(Same problem in the definition of the `close` event apparently, which should have `data-dfn-for="CloseWatcher,HTMLElement,MessagePort"`).

dontcallmedom:
I guess we don't have but could benefit from a system that detects EventTarget interfaces for which is there is no matching known events?

tidoust:
Yes, I was thinking about that as well. I created #1216 to track this.

tidoust:
Closing as addressed. The `cancel` event entry now correctly reports that the event bubbles on `HTMLInputElement`, does not bubble on `HTMLDialogElement`, and also fires at `CloseWatcher` where bubbling does not mean anything. The update was released in v1.11.0 of `@webref/events`.

PR #1220, opened 2024-04-19
Patch `cancel` event to refine bubbling information

tidoust:
As discussed in #1212, the `cancel` event bubbles on `HTMLInputElement`, but it does not bubble on `HTMLDialogElement`. It does not bubble on `CloseWatcher` either but that's because `CloseWatcher` is not part of a bubbling tree.

Events extraction cannot automatically get this nuance, be it only because it only sees `HTMLElement` as target interface for the event, and not `HTMLInputElement` and `HTMLDialogElement`. In practice, it claims that the event bubbles, which is neither wrong nor right.

This patch restricts the `cancel` entry in the HTML events extract to only target `HTMLInputElement`, in order to create the right bubbling entry. And it creates another `cancel` entry which does not bubble for `HTMLDialogElement` and `CloseWatcher`.

With this patch (and the new version of Reffy), we should end up with the following entry in the consolidated `events.json` file, which captures the fact that:
- the event bubbles on `HTMLInputElement`
- the event does not bubble on `HTMLDialogElement`
- the concept of bubbling is meaningless on `CloseWatcher`

```json
{
  "href": "https://html.spec.whatwg.org/multipage/indices.html#event-cancel",
  "src": {
    "format": "summary table",
    "href": "https://html.spec.whatwg.org/multipage/indices.html#event-cancel"
  },
  "type": "cancel",
  "targets": [
    {
      "target": "HTMLInputElement",
      "bubbles": true,
      "bubblingPath": [
        "Node",
        "Document",
        "Window"
      ]
    },
    {
      "target": "CloseWatcher"
    },
    {
      "target": "HTMLDialogElement",
      "bubbles": false
    }
  ],
  "interface": "Event"
}
```

### DIFF
--- a/tools/amend-event-data.js
+++ b/tools/amend-event-data.js
@@ -81,9 +81,9 @@ const patches = {
       pattern: { type: /^(cut|clipboardchange|paste|copy)$/ },
       matched: 4,
       change: {
-	interface: "ClipboardEvent",
+        interface: "ClipboardEvent",
         targets: ["GlobalEventHandlers"],
-	bubbles: true
+        bubbles: true
       }
     }
   ],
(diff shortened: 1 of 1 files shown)


tools/amend-event-data.js
```diff
@@ -183,15 +183,36 @@ const patches = {
...
+          format: "summary table",
+          href: "https://html.spec.whatwg.org/multipage/indices.html#event-cancel"
+        }
+      }
     },
     {
       pattern: { type: "close"},
       matched: 1,
-      change: { targets: ["HTMLDialogElement"   ] }
+      change: { targets: ["CloseWatcher", "HTMLDialogElement", "MessagePort"] }
```
dontcallmedom, line 215: is this meant to be part of that patch?

tidoust: Forgot to mention it, but yes. Not needed to solve `cancel`, but needed to refine the target interfaces of `close`.

dontcallmedom: 1. can you add a comment to that effect?
2. shouldn't MessagePort and CloseWatcher be fixable upstream? or is it that the aggregation can't deal properly with this case?

tidoust: > can you add a comment to that effect?

Where? Or about what? I added a generic comment in the code already a few lines above that one precisely to explain why we have all of these patches for HTML events: to have something more specific than `HTMLElement`, which is the interface that HTML references because it defines the event handler.

> shouldn't MessagePort and CloseWatcher be fixable upstream? or is it that the aggregation can't deal properly with this case?

That has already been fixed upstream with the PR you sent. That's not why this patch exists. As for the other HTML events, the patch is meant to transform `HTMLElement`, here into `HTMLDialogElement`. I'm adding `CloseWatcher` and `MessagePort` because our patching mechanism is "all or nothing". If I don't repeat `CloseWatcher` and `MessagePort`, it effectively removes them, unfixing the fix that was made upstream...

dontcallmedom: Thanks for the explanation and sorry for the noise 😊

dontcallmedom (approved):

PR #1248, opened 2024-06-06
Add patch for saa-non-cookie-storage

dontcallmedom:
Use invalid non-optional dictionary as last argument; see https://github.com/privacycg/saa-non-cookie-storage/issues/30


tidoust (approved):
I thought the idea was to rather patch the base spec to drop the definition of `requestStorageAccess`, but patching the spec that's not doing the "right thing" seems preferable indeed.

dontcallmedom:
patching the original spec was my initial suggestion, but that's when I thought the new spec was proposing a replacement of the method rather than an overload; since the overload is bogus atm, removing it from the new spec made more sense to me.

PR #1261, opened 2024-06-19
Add patch for ed/css/mediaqueries-5.json

tidoust:
Drop duplicate dfn of `<general-enclosed>`

That second patch should fix curation. In other words, CI tests should pass... although note the curation process now takes ~5mn due to https://github.com/w3c/reffy/issues/1599


tidoust:
Seems to work, merging...

PR #1285, opened 2024-07-11
Add interface to `error` event in Web Audio API

tidoust:
Problem raised in:
- https://github.com/WebAudio/web-audio-api/issues/2590 for the actual interface
- https://github.com/WebAudio/web-audio-api/issues/2591 for the syntax

### DIFF
--- a/tools/amend-event-data.js
+++ b/tools/amend-event-data.js
@@ -434,6 +434,13 @@ const patches = {
       pattern: { type: 'audioprocess' },
       matched: 1,
       change: { interface: 'AudioProcessingEvent' }
+    },
+    {
+      // https://github.com/WebAudio/web-audio-api/issues/2591
+      // https://github.com/WebAudio/web-audio-api/issues/2590
+      pattern: { type: 'error' },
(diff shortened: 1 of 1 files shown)


tools/amend-event-data.js
```diff
@@ -434,6 +434,13 @@ const patches = {
       pattern: { type: 'audioprocess' },
       matched: 1,
       change: { interface: 'AudioProcessingEvent' }
+    },
+    {
+      // https://github.com/WebAudio/web-audio-api/issues/2591
+      // https://github.com/WebAudio/web-audio-api/issues/2590
+      pattern: { type: 'error' },
+      matched: 1,
+      change: { interface: 'Event' }
```
dontcallmedom, line 443: I understand https://github.com/WebAudio/web-audio-api/issues/2590 argues that `ErrorEvent` is wrong, but the spec currently says `ErrorEvent` - shouldn't we stick to the spec until this acknowledges by the spec itself?
```suggestion
      change: { interface: 'ErrorEvent' }
```

PR #1298, opened 2024-07-24
Switch to ECMAScript modules

tidoust:
Same types of updates as done in Reffy (https://github.com/w3c/reffy/pull/1628).
This should allow us to resume bumping dependencies that have switched to ESM modules as well and can no longer be imported as CommonJS modules.

Note: another update will be needed to integrate the new ESM version of Reffy, once released, to adjust the code that calls `getSchemaValidationFunction`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   "engines": {
     "node": ">=20"
   },
+  "type": "module",
   "devDependencies": {
     "@actions/core": "1.10.1",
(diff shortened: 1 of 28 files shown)


tidoust:
CI tests fail for unrelated reasons (curation is currently broken due to invalid IDL in a CSS spec)

dontcallmedom (approved):
🙏

PR #1348, opened 2024-09-16
Drop IDL patch for Ink Enhancement

tidoust:
Spec now fixed.


tidoust:
CI tests fail for unrelated reasons

PR #1376, opened 2024-11-02
Add patch for ed/css/css-values-5.json

tidoust:
Amend syntax of `if()`, drop value of `<if-condition>`

Note: This won't be enough for all CI tests to turn green because Turtledove got further updated yesterday to integrate more IDL from Private Aggregation API, leading to new duplicated IDL errors.

### DIFF
--- a/ed/csspatches/css-values-5.json.patch
+++ b/ed/csspatches/css-values-5.json.patch
@@ -3,15 +3,16 @@ From: Francois Daoust <fd@tidoust.net>
 Date: Sat, 2 Nov 2024 10:59:54 +0100
 Subject: [PATCH] Amend syntax of `if()`, drop value of `<if-condition>`
 
-For `if()`, parsing fails on `;?`. That seems like a bug in the CSS parser,
-reported in:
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I completely misread the csstree issues - they were closed because csstree was fixed; do we still need the patches or would an upgrade to csstree suffice?

tidoust:
> I completely misread the csstree issues - they were closed because csstree was fixed; do we still need the patches or would an upgrade to csstree suffice?

Yes, issues have indeed been super promptly addressed in CSSTree and a new version of CSSTree should be all we need... but that new version does not exist yet (version 3.0.1 released a couple of days ago does not include these updates) so we still need to wait a bit. In the meantime, we still need the patches.

Side note: dependabot currently seems to reject any update of css-tree, and I don't understand why for the time being. Logs say "No update possible for css-tree 2.3.1".

 I dropped the links to the issues, and created an internal issue to track the switch and dig into dependabot's logic...

dontcallmedom (approved):

PR #1410, opened 2024-12-17
Make css-gaps supesede css-multicol

tidoust:
Definitions in the new CSS Gaps module supersed those in CSS Multi-column Layout.

Note: CI will continue to fail because the module also re-defines the `repeat()` function, defined in CSS Grid. To be handled through a patch as I'm not clear that any superseding is intended.


tidoust:
Note: integrated in #1411. If you're fine with the patch, I'll close this.

dontcallmedom (approved):

PR #1416, opened 2024-12-17
Adjust clean jobs for CDDL extracts for *real*

tidoust:
Nice try, François, but check should have been done the other way around!


tidoust:
Taking the liberty to merge...

Issue #1417, opened 2024-12-17
Validate CDDL extracts during curation step


tidoust:
Reffy now produces CDDL extracts (see #1353). These CDDL extracts are provided as-is with no guarantee whatsoever. Ideally, we'd be able to provide additional guarantees:

1. Syntax-wise, the CDDL extracts should be valid. This requires making sure that the CDDL can be parsed. [`cddlparser`](https://github.com/tidoust/cddlparser) could perhaps be used but then it's in Python. The main other alternatives are [`cddl` written in Rust](https://github.com/anweiss/cddl) and [`cddlc` written in Ruby](https://github.com/cabo/cddlc).
2. Semantic-wise, the CDDL should "make sense". Additional checks could be done directly by the CDDL parser (e.g. no duplicate rules) and we probably don't need to go much deeper than that. Ideally though, the semantic checks would also consider the combination of CDDL extracts that share the same namespace. This requires a mechanism to tell how CDDL extracts need to be combined. For example, the CDDL defined in WebDriver BiDi is extended by Permissions and Web Bluetooth but these extensions are only [noted informatively in WebDriver BiDi](https://w3c.github.io/webdriver-bidi/#external-specifications). The extension mechanism could perhaps leverage [CDDL Module Structure](https://datatracker.ietf.org/doc/html/draft-ietf-cbor-cddl-modules-03).

tidoust:
> Syntax-wise, the CDDL extracts should be valid.

CDDL syntax is now guaranteed to be valid (according to `cddlparser`) in curated extracts.

PR #1422, opened 2024-12-19
Add a validation guarantee to curated CDDL extracts

tidoust:
This leverages CDDL validation in Strudy to provide a validation guarantee to curated CDDL extracts, see #1417.

A `cddlpatches` folder, which will hopefully remain empty, gets added so that we may temporarily patch CDDL extracts when they become invalid.

The PR is ready for review, but CDDL extracts in Webref are invalid for now. CDDL extracts from the permissions spec should become valid with next crawl. CDDL extracts from the at-driver depend on https://github.com/w3c/at-driver/pull/85. I suggest to leave this PR pending until the extracts in Webref become valid, not to have to start with adding patches.

### DIFF
--- a/test/idl/consistency.js
+++ b/test/idl/consistency.js
@@ -33,7 +33,9 @@ const ignorableAnomalies = [
   'incompatiblePartialIdlExposure',
   'singleEnumValue',
   'unexpectedEventHandler',
-  'wrongCaseEnumValue'
+  'wrongCaseEnumValue',
+  'noEvent',
+  'urlType'
 ];
 
 
--- a/README.md
+++ b/README.md
@@ -92,7 +92,9 @@ Data curation brings the following guarantees.
 
 ### CDDL extracts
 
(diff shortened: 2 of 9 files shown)


test/idl/consistency.js
```diff
@@ -33,7 +33,9 @@ const ignorableAnomalies = [
   'incompatiblePartialIdlExposure',
   'singleEnumValue',
   'unexpectedEventHandler',
-  'wrongCaseEnumValue'
+  'wrongCaseEnumValue',
+  'noEvent',
+  'urlType'
```
dontcallmedom, line 38: I'm not convinced we should add these to the guarantees of webref - this will make the cost of patching higher, for no obvious benefit

dontcallmedom: ha! sorry for the noise then :)

dontcallmedom (approved):

jugglinmike:
I just merged https://github.com/w3c/at-driver/pull/85, so if that's been a blocker here, it's clear now.

dontcallmedom:
the tests now pass \o/

PR #1428, opened 2025-01-06
Bump package-lock.json

tidoust:
No idea why `npm ci` claims that `package.json` and `package-lock.json` have gone out of sync. This is the result of running `npm install` to sync files again.


tidoust:
Tests fail due to curation issues, but at least dependencies get properly installed.

PR #1448, opened 2025-01-23
Auto-drop incorrect wrapping quotes in values

tidoust:
Pending resolution of https://github.com/speced/bikeshed/issues/3011 or updates made to underlying specs, this adds a temporary mechanism to drop offending wrapping single quotes in the definitions of properties and values during data curation, so that we may publish a new version of `@webref/css` while the problem gets fixed.

### DIFF
--- a/tools/drop-css-property-duplicates.js
+++ b/tools/drop-css-property-duplicates.js
@@ -228,6 +228,31 @@ async function dropCSSPropertyDuplicates(folder) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
noting that this stretches the semantics implied by the file name :)

(relatedly, to confirm my understanding - if this wasn't a temporary thing with the associated pain of sync'd releases, it would more logically fit as a reffy post-processing item, right?)

PR #1452, opened 2025-01-29
Re-generate patch for CSS UI

tidoust:
Patch no longer applied due to some unrelated update in the spec.


tidoust:
CI tests fail on invalid IDL. Unrelated.

Issue #1426, opened 2024-12-21
Question: Is it possible to identify possible errors for an operation from the JSON files?


stroiman:
I'm working on a "small" project, a [headless browser in Go](https://github.com/stroiman/go-dom) with an embedded v8; and I've started using this project as a source to generate Go code. So far, primarily v8 wrapper on top of native Go implementations.

For this purpose, I'd very much like to know if each operation can result in an error, particularly because of how idiomatic error handling in Go works.

I am consuming files from `webref/curated/idlparsed`, e.g. for the `URL` type, I consume `webref/curated/idlparsed/url.json` (I assume this is a good source for this use case?).

The URL constructor can throw a `TypeError`, and that information is in prose in `curated/algorithms/url.json`, but I wasn't able to find that information in a more precise manner.

Is that information available in easy-to-consume machine readable format?

tidoust:
> I am consuming files from webref/curated/idlparsed, e.g. for the URL type, I consume webref/curated/idlparsed/url.json (I assume this is a good source for this use case?).

Yes, or the IDL extracts directly from webref/curated/idl if you prefer to use your own Web IDL parser. The curated branch is the right place to look at, in any case!

> Is that information available in easy-to-consume machine readable format?

I'm afraid it isn't.

There is no standard way to express that information in Web IDL. See https://github.com/whatwg/webidl/issues/603 for a proposal to annotate Web IDL definitions that may throw (but it would not give you the type of errors being thrown as far as I can tell), also https://github.com/rustwasm/wasm-bindgen/issues/2983 where you'll find part of the discussion and manually maintained Web IDL fragments annotated with custom extended attributes.

Specifications do define that information in prose. It may be possible to extract that prose and create machine readable extracts out of it. That likely requires a bit of work on an on-going basis (similar to what we do for CSS and IDL extracts), including working with editors to converge on common prose constructs to express the information in the specs. I don't think we're going to jump on it ourselves for lack of time, but if you're interested in digging this up further, we'd certainly be happy to help!

stroiman:
Thanks a lot for taking the time to answer, and expressing being happy to help.

For now, I will just handle this in generator code, having a data structure that can provide options to override default behaviour for specific attributes/operations.

At the moment, I only need a boolean value, _if_ a method can result in an error. The type of error isn't relevant yet.

PR #1456, opened 2025-02-04
Add links to Gost-DOM

stroiman:
I created a headless browser in Go, and use the webref to generate primarily JavaScript bindings to Go objects, as well as mapping element tagnames to IDL interface name.

The actual parsing of webref files were extracted as its own library.

It's very early prototype supporting only a very small subset of the standard, but it does support a session-based username/password login flow using HTMX.

So if it's enough to be in the list must be for you to judge :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ The following projects are known to use webref data:
 * [nodysseus](https://gitlab.com/ulysses.codes/nodysseus)
 * [WebIDLPedia](https://dontcallmedom.github.io/webidlpedia/)
 * [Webdex](https://dontcallmedom.github.io/webdex/)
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
thanks for the link, and glad to see webref used in a new project! please let us know when you see opportunities for improving webref as you make use of it

stroiman:
Awesome :)

I already had a discussions, that I'd really like to know if the specs define if a method can result in an error, but it isn't part of the idl standard, yet. 

But it's not much of an issue, I already need to have the ability to supply customization to how a method should be generated, so adding a `SetNoError()` on some of the methods isn't too much of an issue. 

https://github.com/w3c/webref/issues/1426

PR #1457, opened 2025-02-04
Drop CSS patch for CSS Overflow Level 5

tidoust:
Spec fixed upstream.


tidoust:
Note: CI tests will fail for unrelated reasons.

PR #1486, opened 2025-03-06
Add patch for ed/idl/svg-paths.idl

tidoust:
I'm not sure how to fix the `values` attribute of the `SVGPathSegment` interface. Or rather, I'm not sure how that attribute is going to be fixed in the spec. I propose to drop it in the meantime.

Note CI tests will fail because another patch is needed for the SVG spec to drop `SVGPathElement`.

### DIFF
--- a/ed/idlpatches/svg-paths.idl.patch
+++ b/ed/idlpatches/svg-paths.idl.patch
@@ -1,0 +1,46 @@
+From 9a31ecbbc7706498e54712fa204a0b9d2b7a26cf Mon Sep 17 00:00:00 2001
+From: Francois Daoust <fd@tidoust.net>
+Date: Thu, 6 Mar 2025 08:52:46 +0100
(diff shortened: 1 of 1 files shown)


dontcallmedom:
I'd have an inclination to keep the `values` attribute (either with `FrozenArray`, or falling back to `any`) - I assume this would help with regard to BCD to have all expected attributes in an implementation.

(I'm surprised the [Mozilla WebIDL](https://searchfox.org/mozilla-central/source/dom/webidl/SVGPathElement.webidl) parser isn't having an issue with the `sequence`-typed attribute)

tidoust:
> I'd have an inclination to keep the values attribute (either with FrozenArray, or falling back to any) - I assume this would help with regard to BCD to have all expected attributes in an implementation.

Done using `FrozenArray`.

dontcallmedom (approved):

PR #1487, opened 2025-03-06
Add patch for ed/idl/SVG.idl

tidoust:
Drop SVGPathElement interface


dontcallmedom:
this seems to be losing the content of the existing patch (hence the CI failure?)

PR #1488, opened 2025-03-06
Add patch for ed/idl/SVG.idl

tidoust:
Second try (see #1487)


dontcallmedom (approved):

dontcallmedom:
thinking some more: we probably need an issue on this repo to reference in the patch, otherwise the patch will be deleted if the other SVG issue gets closed (one can always dream)

PR #1571, opened 2025-06-06
Add patch for ed/css/pointer-animations.json

tidoust:
Prefix a couple of type definitions to avoid duplication


tidoust:
Seems there remains another unrelated hiccup (invalid link in css-shapes-2). Will look into it. That's unrelated and the curation error occurs after the check on duplicated types.

dontcallmedom (approved):

PR #1572, opened 2025-06-06
Create logic to publish two versions of `@webref/css`

tidoust:
Goal is to continue to publish the CSS extracts per spec as version 6 and to publish the consolidated CSS as version 7 of the `@webref/css` package.

The versions are handled as two different packages in the code, but the actual package name is the same (defined in respective `package.json` files).

Note: Draft PR for now as I still need to add an `-alpha` suffix to the new version and update the code that bumps the package version accordingly as it does not currently expect that.

### DIFF
--- a/tools/prepare-release.js
+++ b/tools/prepare-release.js
@@ -248,11 +248,14 @@ ${diff.substring(0, 60000)}`;
   console.log("Extract and bump version number");
   const packageFilename = path.resolve(scriptPath, '..', 'packages', type, 'package.json');
   const packageFile = await loadJSON(packageFilename);
-  const version = packageFile.version;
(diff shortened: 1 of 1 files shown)


tidoust:
Code should now properly support pre-release version flags. Proposal is to release a `7.0.0-alpha` version. I also adjusted the pre-release PR script so that it installs and computes a diff with the right version of the released package. 

I tried to document the guarantees and to point people at "How to upgrade" instructions in the documentation.

tidoust:
Note: I couldn't really test the release logic, coz' it's tied to npm releases. I wouldn't be surprised if there were details to adjust. Nothing like testing in production, right? \o/

dontcallmedom (approved):

PR #1588, opened 2025-06-16
Add `index()` function to CSS package

tidoust:
The function creates an indexed structure more aligned with MDN data. The README documentation was also updated to detail remaining structural nuances, and more importantly differences in content between the two projects.


tidoust:
The CI test failure is unrelated, but I still need to understand why it gets raised. As far as I can tell, the problematic type was introduced in `main` **after** that pull request got created and tests should not run against data in `main`.

I do see a minor flaw in the code that applies "freeze" patches (the code assumes that it needs to return to `main` afterwards). But that should not have much consequences on top of leaving the client on the wrong branch afterwards.

PR #1591, opened 2025-06-20
Add patch for ed/css/css-values-5.json

tidoust:
Fix dfn of `<input-position>`. I haven't tried to create an entry for `<output-value>` (not a very interesting type in any case as it does not come with an explicit syntax).


tidoust:
Another CSS syntax hiccup needs to be taken care of... but that's unrelated.

PR #1624, opened 2025-07-24
Adjust `index()` in CSS package to also index descriptors

tidoust:
The consolidated CSS file is mostly flat except for descriptors that are nested under at-rules. The descriptors were still using an array. For consistency, the `index()` function now indexes them by name too.

README was also updated to drop the note on at-rules syntax expansion since we no longer do that. It also notes that the list of descriptors is not guaranteed to be exhaustive.

<del>Draft PR for now because test reveals the need to consolidate descriptors too: some of them get redefined in subsequent levels</del> (Edit: fixed in Reffy).


dontcallmedom:
the schema expects `type` to be set

tidoust:
Doh, I was pretty sure I had looked at the schema!

tidoust:
Schema updated (tests will keep on failing for a few unrelated reasons though)

PR #1670, opened 2025-09-15
Re-generate patch for ed/css/SVG.json

tidoust:
Publication of SVG Editors's Drafts at svgwg.org has resumed.


tidoust:
Wrong patch, need to re-generate the IDL patch instead...

PR #1671, opened 2025-09-15
Re-create IDL patch for SVG

tidoust:
Publication of SVG Editor's Drafts to svgwg.org has resumed. This removes the need to patch inclusion of `DocumentAndElementEventHandlers`.


tidoust:
But this brings at least one additional invalid IDL construct that needs patching as well. Will look into it and re-generate patch.

PR #1672, opened 2025-09-15
Re-geneate IDL patch for SVG

tidoust:
Re-generation re-introduces the patch for the constructor operation.


tidoust:
Yet another construct that needs fixing...

PR #1676, opened 2025-09-22
Re-generate IDL patch for ed/idl/SVG.idl

tidoust:
Editor's Draft was updated... but svgwg.org does not yet return the latest version that includes further fixes. Patch will fail again when the ED gets updated. That's good.


tidoust:
I'll end up getting it right ;)

Issue #1737, opened 2025-12-03
Missing syntax for several CSS types


caugner:
## Background

We have deprecated [mdn-data](https://www.npmjs.com/package/mdn-data), and recommend consumers to migrate to [@webref/css](https://www.npmjs.com/package/@webref/css), but it appears that mdn-data includes some data that is missing in webref. (See also: https://github.com/csstree/csstree/pull/355/files#r2585312764)

## Problem

The following 115 types have no syntax in [`css.json`](https://unpkg.com/@webref/css/css.json):

### https://drafts.csswg.org/css-animations-2/

- [`<animation-action>`](https://drafts.csswg.org/css-animations-2/#typedef-animation-action)

### https://drafts.csswg.org/css-box-4/

- [`<box>`](https://drafts.csswg.org/css-box-4/#typedef-box)

### https://drafts.csswg.org/css-cascade-6/

- [`<scope-end>`](https://drafts.csswg.org/css-cascade-6/#typedef-scope-end)
- [`<scope-start>`](https://drafts.csswg.org/css-cascade-6/#typedef-scope-start)

### https://drafts.csswg.org/css-color-4/

- [`<deprecated-color>`](https://drafts.csswg.org/css-color-4/#typedef-deprecated-color)
- [`<hex-color>`](https://drafts.csswg.org/css-color-4/#typedef-hex-color)
- [`<named-color>`](https://drafts.csswg.org/css-color-4/#typedef-named-color)
- [`<quirky-color>`](https://drafts.csswg.org/css-color-4/#typedef-quirky-color)
- [`<system-color>`](https://drafts.csswg.org/css-color-4/#typedef-system-color)

### https://drafts.csswg.org/css-conditional-5/

- [`<boolean-condition>`](https://drafts.csswg.org/css-conditional-5/#typedef-boolean-condition)
- [`<scroll-state-feature>`](https://drafts.csswg.org/css-conditional-5/#typedef-scroll-state-feature)
- [`<size-feature>`](https://drafts.csswg.org/css-conditional-5/#typedef-size-feature)
- [`<style-feature-name>`](https://drafts.csswg.org/css-conditional-5/#typedef-style-feature-name)
- [`<style-feature-value>`](https://drafts.csswg.org/css-conditional-5/#typedef-style-feature-value)
- [`<supports-condition-name>`](https://drafts.csswg.org/css-conditional-5/#typedef-supports-condition-name)

### https://drafts.csswg.org/css-counter-styles-3/

[...]

Issue #1647, opened 2025-08-20
Add formal syntaxes for `<system-color>` and `<deprecated-color>`


tidoust:
Via https://github.com/w3c/webref/issues/1519#issuecomment-3120290261

The CSS color specs define a number of values that the `<system-color>` and `<deprecated-color>` types can take but not synthesize things in a formal syntax. Ideally, the syntax of these types would be:

```
<system-color> = AccentColor | AccentColorText | [...] | <deprecated-color>
<deprecated-color> = ActiveBorder | ActiveCaption | [...]
```

Approaches to consider to get these syntaxes into Webref by order of preference:
1. Craft a PR against the specs to introduce the formal syntax there.
2. Explore the possibility to extract and use values in Reffy when a more formal syntax does not exist. Make sure that this does not create bad syntaxes for other constructs though (and note a patch would likely still be needed to add `<deprecated-color>` to `<system-color>`).
3. Create a patch in Webref, but note the patching mechanism is not well-suited for this type of permanent updates.

cdoublev:
(1) has been rejected a few times in the past, like for `<dimension-unit>` (now `<attr-unit>`) in https://github.com/w3c/csswg-drafts/issues/9648#issuecomment-1832692281. But I think [`<absolute-size>` and `<relative-size>`](https://drafts.csswg.org/css-fonts-4/#font-size-prop), [`<age>`](https://drafts.csswg.org/css-speech-1/#typedef-voice-family-age), [`<gender>`](https://drafts.csswg.org/css-speech-1/#typedef-voice-family-gender), could be updated.

I explored (2) but quickly dismissed it because the logic to identify these keyword lists would certainly be brittle.

Below are the productions that expand to a keyword list but that I have not mentioned above:

  - `<named-color>`
  - `<outline-line-style>`
  - `<page-size>`
  - `<scroll-button-direction>`
  - `<timeline-range-center-subject>`
  - `<timeline-range-name>`

I hope this helps! =)

cdoublev:
I would help reporting any update if you decide to produce this data, but I do not think it is worth it.

That said, if specs do not output a formal syntax because it would not be human-readable, `bikeshed` should ideally produce a markup for machines to synthesize it, assuming it cannot do it by itself (in a *Type Index* appendice).

For example, with an `alternative-for` attribute assigned the CSS type producing it. This would allow to synthetize a CSS type from alternatives defined in different specs, like `<color>` or `<timeline-range-name>`.

[...]

PR #1756, opened 2025-12-16
Add patch mechanism to add CSS syntaxes manually

tidoust:
This is meant to provide a simple patching mechanism for #1647 and #1737, based on a JSON-like structure, similar to the patching mechanism that we use for events (`amend-event-data.js`).

Patches get applied during curation after dropping duplicate definitions, and before the CSS extracts get consolidated.

The patching logic is initially restricted to adding syntaxes where they do not exist (no way to override an existing one for now) and to adding syntaxes to base definitions (no way to set the syntax of a `newValues` property).

The logic allows to say:
1. Here is the syntax, period. That's the most direct way. But also the one that requires manual maintenance.
2. Compute the syntax from the list of values, and optionally complete that list with this additional syntax. This allows to compute syntaxes such as `<system-color>` indirectly but still automatically from the spec.
3. Use the syntax from that other construct (must be defined in the same spec). This allows to say that the syntax of `-webkit-user-select` is the same as that of `user-select`.

It will probably be useful to extend the patching logic slightly to allow overriding existing syntaxes over time. This would allow us to replace most patches in `csspatches` with a more convenient mechanism.

The initial list of patches is based on missing syntaxes identified in the mdn-webref analysis:
https://github.com/tidoust/mdn-webref/blob/main/report-syntax.md#syntax-mismatches-between-mdn-data-and-webref ... completed with some of the types highlighted in #1647.

At-rules and selectors are excluded for now. They can be patched too, just not done in that iteration. A few other ones are missing when it was not obvious to me that the syntax was correct, or when it seemed that the underlying spec could perhaps be updated.

In other words, the initial list of patches should be a good start but is not meant to be exhaustive.

### DIFF
--- a/ed/csspatches/syntax-patches.js
+++ b/ed/csspatches/syntax-patches.js
@@ -143,7 +143,7 @@ export default {
       <rotate3d()> <rotate()> | <rotateX()> | <rotateY()> | <rotateZ()> |
       <skew()> | <skewX()> | <skewY()> |
       <matrix3d()> | <matrix()> | <perspective()>
-    `,
+    `
   },
 
   // https://drafts.csswg.org/css-ui-4/#propdef--webkit-user-select


dontcallmedom:
Looks great, thanks!

Issue #1851, opened 2026-02-28
Incorrect and missing CSS `longhands` entries


cdoublev:
The `longhands` property has recently been added for shorthands in `@webref/css`, but there are two problems:

  - it includes entries that are not longhands but shorthand sub-properties
  - it is missing reset-only sub-properties

For example, the `longhands` of `border` are defined with `border-width`, `border-style`, `border-color`. I do not think that a definition of *"longhand*" exists in CSS specs, but CSS Cascade 5 defines [this](https://drafts.csswg.org/css-cascade-5/#shorthand-property):

  > A shorthand property sets all of its longhand sub-properties, exactly as if expanded in place.

Ie. a shorthand sets a declaration for all its longhands. But there is no shorthand declarations internally stored.

So I believe that these are *shorthand sub-properties*, not longhands.

The `longhands` of `border` are missing `border-image-*` longhands, which are its reset-only sub-properties. When `border` is declared, it sets a declaration for all its longhands, including its reset-only sub-properties.

If you are interested, you can see my own maintained version of the shorthand/longhands mapping [here](https://github.com/cdoublev/css/blob/master/lib/properties/shorthands.js). Note that I split reset-only subproperties in a separate group, which would be handy in `@webref/css`.

cdoublev:
When you declare `border` using [`CSSStyleDeclaration.setProperty()`](https://drafts.csswg.org/cssom-1/#dom-cssstyledeclaration-setproperty), the spec says:

  > 8. If `property` is a shorthand property, then for each longhand property `longhand` that `property` maps to, in canonical order, follow these substeps:
  >     1. Let `longhand result` be the result of [set the CSS declaration](https://drafts.csswg.org/cssom-1/#set-a-css-declaration) `longhand` with the appropriate value(s) from `component value list` [...]

If you think implementations should set a declaration for `border-width`, `border-style`, `border-color`, please take a look at the algorithm for `CSSStyleDeclaration.getPropertyValue()`, to understand why this would not work.

The _canonical order_ is defined in the property definition table, but it is always *"per grammar"* (which is not enough for some shorthand, but I disgress). For `border`, the grammar is `<line-width> || <line-style> || <color>`. So expanding to the `border-<side>` shorthands is misleading *and* incorrect, fair enough.

---

[...]

PR #1872, opened 2026-03-11
Add a `resetLonghands` field to CSS shorthand properties

tidoust:
Via #1851. Some shorthand properties may also reset a number of properties that they don't set. These properties are known as "reset-only sub-properties": https://drafts.csswg.org/css-cascade-5/#reset-only-sub-property

This update adds a `resetLonghands` field in CSS extracts to shorthands that have reset-only sub-properties, leveraging the list maintained by @cdoublev in:
https://github.com/cdoublev/css/blob/master/lib/properties/shorthands.js

Affected properties:
- `animation` (and the legacy `-webkit-animation`)
- `background`
- `border`
- `font`
- `mask` (and the legacy `-webkit-mask`)

The list needs to be maintained manually, I'm afraid. I don't see any good way to extract the list of reset-only sub-properties automatically for now.

Note: I haven't listed `animation-composition` as a reset-only sub-property for `animation` yet, pending clarity on its inclusion, see https://github.com/w3c/webref/issues/1851#issuecomment-4040943080. To be determined before this PR is merged!

### DIFF
--- a/tools/add-css-longhands.js
+++ b/tools/add-css-longhands.js
@@ -376,6 +376,41 @@ const MANUAL_EXPANSIONS = {
 
 
 /**
+ * List of reset only sub-properties for shorthands. URLs in comments link to
+ * the spec sections that make that relationship explicit.
+ */
+const RESET_LONGHANDS = {
+  // https://drafts.csswg.org/scroll-animations-1/#named-range-animation-declaration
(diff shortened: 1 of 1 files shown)


dontcallmedom (approved):
LGTM, with a potential improvement (likely for a separate PR)

Issue #1874, opened 2026-03-16
Document extract fields


tidoust:
Via #1851 and a few other past exchanges. Core fields in CSS extracts are directly taken from [CSS Property definition conventions](https://www.w3.org/TR/CSS2/about.html#property-defs), but we keep adding new fields and do not document their meaning and scope anywhere, leaving it to consumers to infer whatever they feel is the right interpretation.

Extracts would benefit from a bit of documentation in practice:
- [x] CSS (done in #1880)
- [ ] Definitions
- [ ] Elements
- [ ] Events
- [ ] Headings, etc.
- [ ] IDs
- [ ] Links
- [ ] Refs
- [ ] And parsed IDL extracts

Issue #1519, opened 2025-04-08
Group CSS features (and `@webref/css` v7 alpha release)


nzakas:
One of the valuable parts of the [`mdn-data`]() package is how it separates CSS features into different categories:

- At-rules
- functions
- properties
- selectors
- syntaxes
- types
- units

In the current webref package, it's just a collection of objects that we then need to dig into to figure out what types are contained within. It would be helpful if the categories could be exposed at the top level of the package and list every entry for that category regardless of spec.

tidoust:
I note the current `@webref/css` package already separates at the root level between:
* at-rules
* properties
* selectors
* and "values", which is a mixed bag of things.

The mixed bag of things exists because CSS specs do not really distinguish between other types when they define concepts. There is a notion of `function` but the specs do not necessarily use that consistently. That ambiguity seems to appear in mdn-data too. For example, the `abs()` function appears both as a "function" and as a "syntax" in mdn-data.

CSS specs do use a `type` definition type too, which could perhaps be used to populate a related category. There seems to be many more `type` definitions in specs than in what mdn-data currently lists as types. For example, `line-color-list`, `linear-color-stop`, `ident-token` are all type definitions from a spec perspective. If they are not in the list on purpose, is there a way to distinguish between types?

CSS specs define units as `value` definitions that are for something. It may be relatively easy to assemble the list of units automatically with a short list of underlyling types. For example, looking at all values defined for `<angle>`, `<length>` and a few others.

Essentially, the question is: can CSS features be categorized automatically? If not, what amount of manual data would need to be maintained?

nzakas:
Thanks for the response. A follow-up question: assuming everyone wants `webref` packages to be as useful as possible, is there a reason the specs themselves can't be updated to encode this information where appropriate?

nzakas:
Ah gotcha, thanks for explaining. :+1:

tidoust:
I explored a bit the differences between MDN data and Webref, see underlying code in [`tidoust/mdn-webref`](https://github.com/tidoust/mdn-webref/), along with the results:

1. The [`webref.json`](https://raw.githubusercontent.com/tidoust/mdn-webref/refs/heads/main/webref.json) file, which could represent what we may want to end up with in Webref to ease consumption of data.
2. The [report](https://github.com/tidoust/mdn-webref/blob/main/report.md), which highlights differences between the two projects.

[...]

PR #1880, opened 2026-03-24
Document the consolidated CSS file

tidoust:
This is a re-write of the `@webref/css` package README to document the JSON structure more thoroughly. All keys are now described with references to a definition in CSS specs when possible, considerations about their meaning and what to expect from key values, along with examples.

The README also includes example logic to gather information about specs from their URLs using browser-specs, as discussed in:
https://github.com/w3c/webref/issues/1519#issuecomment-3407080809

There are a couple of cases where it may be valuable to clean up the data (that would introduce breaking changes though):
- The `extended` key is always set, but then it's only meaningful for properties. It might become useful for other types of CSS features in the future, but could be dropped in the meantime not to create false expectations.
- Descriptors have a `for` key that targets their parent at-rule. That sort of goes without saying. I haven't documented the key, especially since the `for` key is an array in all other cases.

There may also be a couple of cases where we could strengthen guarantees without necessarily creating more work for us, e.g., to clean up `initial` and `animationType` values. To be addressed separately!

Via #1874.

### DIFF
--- a/packages/css/README.md
+++ b/packages/css/README.md
@@ -54,7 +54,7 @@ The following keys are always set for at-rules:
 - [`name`](#name)
 
 Additional keys may be set:
-- [`for`](#syntax)
+- [`for`](#for)
 - [`prose`](#prose)
 - [`syntax`](#syntax)
 


packages/css/README.md
```diff
@@ -4,16 +4,18 @@ This package contains a consolidated list of CSS features defined across specs,
...
 
-# API
+## Usage
 
 Two async methods are exposed: `listAll()` and `index()`. They both resolve with an object that lists CSS features by feature type: `atrules`, `functions`, `properties`, `selectors` and `types`.
 
-The difference between the two functions is that, under each feature type, `listAll()` returns an array of features, whereas `index()` returns an object indexed by feature name to ease direct lookups (and [transition from `mdn/data`](#transition-from-mdndata). Both functions return the same features.
+The difference between the two functions is that, under each feature type, `listAll()` returns an array of features, whereas `index()` returns an object indexed by feature name to ease direct lookups (and [transition from `mdn/data`](#transition-from-mdndata)). Both functions return the same features.
+
+*Note:* The NPM package targets JavaScript runtimes more directly but the data appears in a standalone `css.json` file within the package. This file may be parsed as JSON in any programming language to yield the same object as that returned by the `listAll()` method.
```
dontcallmedom:
```suggestion
[!NOTE]
The NPM package targets JavaScript runtimes more directly but the data appears in a standalone `css.json` file within the package. This file may be parsed as JSON in any programming language to yield the same object as that returned by the `listAll()` method.
```

dontcallmedom (approved):
Great addition, thanks! LGTM with a few nits/suggestions

PR #1907, opened 2026-04-16
Security: Shell command injection risk from package directory name in execSync

tomaioo:
## Summary

Security: Shell command injection risk from package directory name in execSync

## Problem

**Severity**: `High` | **File**: `tools/bump-packages-minor.js:L54`

The script interpolates `type` into a shell command: ``execSync(`git ls-files --others --deleted --exclude-standard --directory ed/${type}`)``. `type` comes from directory names under `packages`, which can be influenced by repository content. A crafted folder name containing shell metacharacters could execute arbitrary commands when this script runs.

## Solution

Use `execFileSync('git', ['ls-files', '--others', '--deleted', '--exclude-standard', '--directory', `ed/${type}`])` instead of shell strings. Also validate `type` against expected package names (e.g., `['css','css6','elements','events','idl']`).

## Changes

- `tools/bump-packages-minor.js` (modified)

### DIFF
--- a/tools/bump-packages-minor.js
+++ b/tools/bump-packages-minor.js
@@ -16,11 +16,16 @@
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { execSync } from 'node:child_process';
+import { execFileSync } from 'node:child_process';
 import { loadJSON } from './utils.js';
 const scriptPath = path.dirname(fileURLToPath(import.meta.url));
+const validPackageTypes = ['css', 'css6', 'elements', 'events', 'idl'];
 
 async function checkPackage(type) {
+  if (!validPackageTypes.includes(type)) {
+    throw new Error(`Unknown package type: ${type}`);
+  }
+
   console.log(`Check ${type} package`);
   const packageFile = path.resolve(scriptPath, '..', 'packages', type, 'package.json');
   const packageContents = await loadJSON(packageFile);
@@ -42,8 +47,8 @@ async function checkPackage(type) {
     return;
   }
 
-  const res = execSync(
-    `git ls-files --others --deleted --exclude-standard --directory ed/${type}`,
+  const res = execFileSync('git',
+    ['ls-files', '--others', '--deleted', '--exclude-standard', '--directory', `ed/${type}`],
     { encoding: 'utf8' }).trim();
   if (res) {
     console.log('- new/deleted files found');


nberlette:
While this is indeed a security risk, is there any real viability to it, given that the file in question is not a part of any public API and is never exposed to userland code or input?

Maybe a maintainer could chime in here (and please correct me if I'm mistaken!), but my initial read on this is that code in the `tools/` folder is only ever going to be run intentionally, without any untrusted inputs... I can't visualize a scenario where the folder names that are interpolated in the `execSync` call site are able to be manipulated by a bad actor ahead of time. However, if this code is executed automatically in, for example, response to any new PR being created or updated, that would introduce a potential attack vector as you've described.

tomaioo:
Great question—you're right that this isn’t a public API/input path, so the risk depends on **who can influence repo contents** where the script runs. The viable scenario is CI or a maintainer running the tool on a branch/PR (including forks) where a crafted package directory name is present; in that context, repo files are effectively untrusted input. So this is mostly a hardening/supply-chain fix: low likelihood in normal use, but high impact if triggered, and `execFileSync` + allowlisting removes the class of issue with minimal cost.

tomaioo:
Great question—you're right that this isn’t a public API/input path, so the risk depends on **who can influence repo contents** where the script runs. The viable scenario is CI or a maintainer running the tool on a branch/PR (including forks) where a crafted package directory name is present; in that context, repo files are effectively untrusted input. So this is mostly a hardening/supply-chain fix: low likelihood in normal use, but high impact if triggered, and `execFileSync` + allowlisting removes the class of issue with minimal cost.

tomaioo:
Great question—you're right that this isn’t a public API/input path, so the risk depends on **who can influence repo contents** where the script runs. The viable scenario is CI or a maintainer running the tool on a branch/PR (including forks) where a crafted package directory name is present; in that context, repo files are effectively untrusted input. So this is mostly a hardening/supply-chain fix: low likelihood in normal use, but high impact if triggered, and `execFileSync` + allowlisting removes the class of issue with minimal cost.

tools/bump-packages-minor.js
```diff
@@ -16,11 +16,16 @@
 import fs from 'node:fs/promises';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { execSync } from 'node:child_process';
+import { execFileSync } from 'node:child_process';
 import { loadJSON } from './utils.js';
 const scriptPath = path.dirname(fileURLToPath(import.meta.url));
+const validPackageTypes = ['css', 'css6', 'elements', 'events', 'idl'];
```
tidoust, line 22: Hardcoding the list of valid package types seems wrong. A simple regular expression check that validates that the type is a simple word (something like `/^[\w-]+$/`) would seem enough to avoid any injection possibility.

(Also, it's not clear to me that sanitizing inputs is actually needed if `execFileSync` gets used (that only seems needed if the `shell` option is enabled)

tidoust (changes requested):
Yeah, the likelihood that this ever turns into a concrete issue seems extremely thin but I guess the same kind of tools that get used to detect and report these potential injection risks could also be used to create artificial PRs that look legit but sneak in a package name like `muahahaha; rm -rf /` that we could perhaps miss on a bad day.

PR #1901, opened 2026-04-08
feat: add ESM support, type definitions, and sync APIs

nberlette:
Closes #1900. Please refer to the tracking issue for the motivation and changes in this PR.

If there's any changes I should make, feel free to let me know either on the associated issue, or right here on the PR itself.

Looking forward to any and all feedback. Thanks!

PS: I realize I've combined a few unrelated changes here into one PR... if that's a problem, I'm happy to split this into 2-3 separate pulls focused on each set of changes.

PR #1908, opened 2026-04-16
feat(types): add type declarations to all `@webref/*` packages

nberlette:
Closes #1900.

## Overview

This PR contains just the type-related changes from the original draft PR #1901, applied to the latest commit on the main branch. In addition to new `index.d.ts` files, each `package.json` manifest has also been updated with a new `types` and `exports` fields.

## Changes

- feat(types): add type declarations for @webref/css package
- feat(types): update @webref/css package manifest
- feat(types): add types to @webref/css6 package
- feat(types): add type declarations for @webref/elements package
- feat(types): add type declarations for @webref/events package
- feat(types): add type declarations for @webref/idl package

### DIFF
--- a/packages/css/index.d.ts
+++ b/packages/css/index.d.ts
@@ -98,9 +98,9 @@ export interface CssAtRuleDefinition extends ScopedCssFeatureDefinition {
 }
 
 /**
- * Indexed variant of a CSS at-rule definition returned by {@linkcode index} and
- * {@linkcode indexSync}. The descriptor array is replaced with an object keyed
- * by descriptor name for direct lookups.
+ * Indexed variant of a CSS at-rule definition returned by {@linkcode index}.
+ * The descriptor array is replaced with an object keyed by descriptor name for
+ * direct lookups.
  */
 export interface IndexedCssAtRuleDefinition
   extends Omit<CssAtRuleDefinition, "descriptors"> {
@@ -202,8 +202,7 @@ export interface CssPropertyDefinition extends CssFeatureDefinition {
 }
 
 /**
- * Shape of the consolidated CSS data returned by {@linkcode listAll} and
- * {@linkcode listAllSync}.
+ * Shape of the consolidated CSS data returned by {@linkcode listAll}.
  */
 export interface CssDefinitions {
   /**
@@ -229,8 +228,7 @@ export interface CssDefinitions {
 }
 
 /**
- * Shape of the indexed CSS data returned by {@linkcode index} and
- * {@linkcode indexSync}.
+ * Shape of the indexed CSS data returned by {@linkcode index}.
  */
 export interface IndexedCssDefinitions {
   /**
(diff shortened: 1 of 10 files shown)


nberlette:
Whoops! I accidentally forgot to trim the synchronous APIs from the type declarations - I'm patching them now and verifying the types accurately reflect what each package actually exposes to the runtime.

packages/css/index.d.ts
```diff
@@ -0,0 +1,282 @@
...
+  readonly canonicalOrder?: string | undefined;
+  /**
+   * Computed value metadata from the property definition table.
+   */
+  readonly computedValue?: string | undefined;
+  /**
+   * Legacy raw key copied from some sources where the field name includes a
+   * non-breaking space.
+   */
+  readonly "computed\u00A0value"?: string | undefined;
```
tidoust, line 161: Ah, this reveals a bug in the extraction logic, which fails to turn that into `computedValue`. That seems to come from the SVG spec. I'll look into fixing that.

packages/css/index.d.ts
```diff
@@ -0,0 +1,282 @@
...
+   */
+  readonly resetLonghands?: string[] | undefined;
+  /**
+   * CSSOM attribute names generated for this property.
+   */
+  readonly styleDeclaration: string[];
+  /**
+   * Used value metadata when exposed by the source specification.
+   */
+  readonly usedValue?: string | undefined;
```
tidoust, line 201: No need to change anything, just noting that this one only seems to appear once in the data, in the definition of [`transform-style`](https://drafts.csswg.org/css-transforms-2/#propdef-transform-style).

tidoust (changes requested):
Thanks! One thing that's currently missing is a test for each package that makes sure that the data in the package actually aligns with these types.

Another thing that would seem useful to me would be to automate the creation of some of these types, leveraging the underlying JSON schemas in Reffy. The entry points are:
- [Schema of the consolidated CSS file](https://github.com/w3c/reffy/blob/main/schemas/postprocessing/css.json)
- [Schema of individual CSS extracts](https://github.com/w3c/reffy/blob/main/schemas/files/extracts/css.json)
- [Schema of element extracts](https://github.com/w3c/reffy/blob/main/schemas/files/extracts/elements.json)
- [Schema of the consolidated events file](https://github.com/w3c/reffy/blob/main/schemas/postprocessing/events.json)
(There is no schema for IDL since the extracts are raw text file)

There seem to be tools that can convert JSON schemas to TypeScript types, such as [`json-schema-to-typescript`](https://www.npmjs.com/package/json-schema-to-typescript).

Note: the schema for the consolidated CSS file is incomplete in Reffy and allows additional keys. This exercise reveals that it would be useful to complete the schema and review new keys when they get introduced, be it only to trap typos such as "computed value".

FYI, we're going to drop support for the `css6` package soon-ish, it may not be worth creating types for it.

packages/css/index.d.ts
nberlette: Since you've fixed this, I'll go ahead and remove this field entirely

nberlette:
@tidoust Thanks for the feedback! Those are all great suggestions.

I'll write unit tests today to ensure all of the types align correctly with the actual runtime data the packages contain. This should help with alerting maintainers of potential package drift in the future.

I can also automate the generation of the types themselves, but I'm not sure if I'll also be able to preserve all of the field-level documentation comments.. I'll look into it :)

PR #1938, opened 2026-05-07
Drop syntax patch for `<basic-shape>`

tidoust:
Syntax was now added to the spec, although note that `<shape()>` is still missing for now, see https://github.com/w3c/csswg-drafts/pull/13894


tidoust:
CI failure is unrelated (another CSS syntax needs fixing)

PR #1950, opened 2026-05-20
`@webref/xref`: Support lookup of non percent-encoded URLs

tidoust:
Via https://github.com/mdn/browser-compat-data/pull/23958#issuecomment-4495384695

BCD does not use URL hashes that are percent-encoded. Webref does. It seems ok to support both forms in the lookup function.

Note: this creates more cases where the `lookup()` may throw an `Error`. We may want to create a better-looking `ParameterError` at some point to make it clear to caller that these errors are due to a problem on their part. Keeping things simple for now.

This update also fixes the code that indexes URLs from `alternateIds` to percent-encode the IDs, although note that the fix is hypothetical for the time being as there aren't any alternate ID that contains characters that need to be encoded.

### DIFF
--- a/packages-auto/xref/index.js
+++ b/packages-auto/xref/index.js
@@ -63,7 +63,7 @@ function setupExtracts(type, rootFolder) {
         if (entry.alternateIds) {
           for (const id of entry.alternateIds) {
             const url = new URL(entry.href);
-            url.hash = id;
+            url.hash = encodeURIComponent(id);
             const href = url.toString();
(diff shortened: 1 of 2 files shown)


dontcallmedom (approved):

tidoust:
Just occurred to me that `decodeURIComponent` fails when given a hash that has a `%` character not followed by digits. I adjusted the code to consider that the hash is already decoded when that happens.

PR #1953, opened 2026-05-20
`@webref/xref`: Add specs and a few initial lookup options

tidoust:
Via https://github.com/mdn/browser-compat-data/pull/23958#issuecomment-4495384695

Any specific lookup rule requires shipping web-specs within the package, which is what this update does. To make sure that the list matches the data, the list of specs is rather built from the crawl index.

The `lookup()` function may now take an optional object with lookup options as second argument. The following options are supported:

- `standing`: only keep results from specs that have the specific standing. Standing should be `"good"`, `"pending"` or `"discontinued"` (most discontinued specs aren't crawled, but some still are for historical reasons).
- `version`: only keep results from the nightly or release version of the specs. Must be `"nightly"` or `"release"`.
- `series`: accept series URLs. Internally, the URL will only match if the fragment exists in the spec that is known to be the current spec in the series.

*Note:* the code does not yet provide any way to access the list of specs itself. That's another thing that could be added later on if the need arises. For example, this would be useful for callers should they need to create custom filtering logic.

### DIFF
--- a/packages-auto/xref/index.js
+++ b/packages-auto/xref/index.js
@@ -9,7 +9,8 @@ const scriptPath = path.dirname(fileURLToPath(import.meta.url));
  */
 const xref = {
   dfns: [],
-  headings: []
+  headings: [],
+  specs: []
 };
 
 /**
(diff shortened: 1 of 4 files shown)


tools/release-autopackages.js
```diff
@@ -46,7 +72,10 @@ async function releaseXrefPackage(minorBump) {
       const srcFiles = await fs.readdir(srcDir);
       for (const file of srcFiles) {
         if (file.endsWith(`.json`)) {
-          await fs.copyFile(path.join(srcDir, file), path.join(dstDir, file));
+          const spec = specs.find(s => `${s.shortname}.json` === file);
```
dontcallmedom: a very minor nit, but any reason to create that intermediary variable?

packages-auto/xref/index.js
```diff
@@ -125,8 +141,40 @@ export function lookup(url) {
...
   }
   parsedUrl.hash = encodeURIComponent(decodedHash.replace(/^#/, ''));
-  const lookupUrl = parsedUrl.toString();
+  let lookupUrl = parsedUrl.toString();
 
-  const res = urlIndex[lookupUrl.toString()];
-  return structuredClone(res) ?? [];
+  // If the series flag is set, the provided URL may also be a spec series URL,
+  // which we need to convert to the URL of the current spec in the series.
+  if (series) {
```
dontcallmedom: this can be skipped if there is a match in the index, can't it?
```suggestion
  if (!urlIndex[lookupUrl] && series) {
```

dontcallmedom (approved):
LGTM with a couple of nits

PR #1996, opened 2026-07-23
Add CSS syntax patch for `<view-transition-type>`

### LINKED ISSUE #1986: Add CSS syntax patch for: `<view-transition-type>`
### Link to the definition in the spec

https://drafts.csswg.org/css-view-transitions-2/#types-cross-doc

### Syntax of the CSS construct

```code
`<custom-ident>`
```

### Notes

> where `<view-transition-type>` is any `<custom-ident>` other than `none`, and not starting with the string `"-ua-"`.

DevOlabode:
## Summary
- Add a CSS syntax patch so `<view-transition-type>` gets an explicit `<custom-ident>` value during curation.
- The type is defined only in prose in css-view-transitions-2 (any `<custom-ident>` other than `none`, and not starting with `-ua-`), so Reffy cannot extract a formal syntax.

Fixes #1986.

## Test plan
- [x] `npm run curate` applies the `css-view-transitions-2` syntax patch
- [x] Curated `css-view-transitions.json` has `"value": "<custom-ident>"` on `<view-transition-type>`
- [ ] CI curate + test jobs pass on the PR

### DIFF
--- a/ed/csspatches/syntax-patches.js
+++ b/ed/csspatches/syntax-patches.js
@@ -160,6 +160,11 @@ export default {
     "<integer>": "<number-token>"
   },
 
+  // https://drafts.csswg.org/css-view-transitions-2/#typedef-view-transition-type
+  "css-view-transitions-2": {
+    "<view-transition-type>": "<custom-ident>"
+  },
+
   // https://drafts.csswg.org/css2/#value-def-shape
   "CSS2": {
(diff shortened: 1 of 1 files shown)


tidoust (approved):
Thanks! Looks good to me (and no way to be more specific in the syntax to match the additional constraints on `<custom-ident>` defined in the spec).

CI tests fail because the PR is based on a previous commit of the `main` branch that contains raw data that cannot be curated, but that's unrelated.

Issue #1934, opened 2026-05-06
[curation] Consider creating back references views per spec


tidoust:
Via https://github.com/dontcallmedom/webdex/issues/22, https://github.com/whatwg/streams/pull/1369#issuecomment-4291574446. Also https://github.com/speced/bikeshed/issues/3155

Reusing code from Webdex, we could build a file per spec that contains, for each term that the spec defines, the list of specs that references it.

This could perhaps be extended to all IDs, regardless of whether they are actual definitions (e.g., headings).

DevOlabode:
## Proposed approach for per-spec back references

Draft for https://github.com/w3c/webref/issues/1934 — implemented in this PR for review.

### Goal

For each term a specification **defines**, list the other specifications that **reference** it (the Streams editor use case from https://github.com/whatwg/streams/pull/1369#issuecomment-4291574446).

### Scope (v1)

- Public `dfns` only (`access !== "private"`, skip `type === "argument"`), matching [Webdex](https://github.com/dontcallmedom/webdex/blob/main/_tools/generate-pages.js).
- Nightly / curated `ed` data only.
- Only terms with ≥1 **external** referrer (self-references excluded).
- Derived curated view (`tools/prepare-backrefs.js`), not a Reffy crawl extract.
- Deferred: all IDs / headings; NPM package; TR.

### Schema

`ed/backrefs/[shortname].json` on the curated branch:

```json
{
  "spec": { "title": "Streams Standard", "url": "https://streams.spec.whatwg.org/" },
  "backrefs": [
    {
      "id": "readablestream",
      "href": "https://streams.spec.whatwg.org/#readablestream",
      "linkingText": ["ReadableStream"],
      "type": "interface",
      "referencedBy": [
        { "shortname": "fetch", "title": "Fetch Standard", "url": "https://fetch.spec.whatwg.org/" }
      ]
    }
  ]
}
```

### Algorithm

1. Index public dfns by `href` (plus HTML/ES multipage ↔ single-page aliases).
2. Expand each referring spec’s `links` (`rawlinks` + `autolinks`) into `url#anchor` values.
3. Match fragment URLs to dfns; accumulate referrers per defining term.
4. Write one file per defining spec that has at least one referenced term; point `index.json` at it.

Feedback welcome on folder name, `referencedBy` payload, and whether this should eventually move into Reffy as a `post` module.

Implemented in a PR from fork `DevOlabode:curate/create-back-reference-views` for review.

tidoust:
[...]

PR #1997, opened 2026-07-23
Add curated per-spec backrefs extracts

DevOlabode:
## Summary
- Add `tools/prepare-backrefs.js` to build per-spec back-reference extracts from curated `dfns` + `links` (Webdex invert-links approach).
- Wire it into `npm run curate` so `curated/backrefs/[shortname].json` lists, for each publicly defined term with external inbound links, which specs reference it.
- Document the new curated extract and add tests (including Fetch → Streams `ReadableStream`).

Design notes and schema: see `docs/backrefs-proposal.md`.

Related to #1934.

## Test plan
- [x] `node tools/prepare-backrefs.js curated` writes extracts (294 files in local run)
- [x] `node --test test/backrefs/all.js` passes
- [x] Streams backrefs include Fetch for `ReadableStream`
- [ ] CI `npm run curate` + `npm test` on the PR


docs/backrefs-proposal.md
```diff
@@ -0,0 +1,45 @@
+## Proposed approach for per-spec back references
```
tidoust: This file can be dropped, no need to detail the proposed approach as a doc file in the repository. The PR and underlying issue already has the discussion.

It would be more interesting to create a sub-section in the main `README.md` under [Curation guarantees](https://github.com/w3c/webref/blob/main/README.md#curation-guarantees) that describes what the backrefs extracts provide (and perhaps documents the schema, pending a better place, which would typically be the README of the npm package if we generate one in the end).

docs/backrefs-proposal.md
```diff
@@ -0,0 +1,45 @@
...
+
+Draft for https://github.com/w3c/webref/issues/1934 — implemented in this PR for review.
+
+### Goal
+
+For each term a specification **defines**, list the other specifications that **reference** it (the Streams editor use case from https://github.com/whatwg/streams/pull/1369#issuecomment-4291574446).
+
+### Scope (v1)
+
+- Public `dfns` only (`access !== "private"`, skip `type === "argument"`), matching [Webdex](https://github.com/dontcallmedom/webdex/blob/main/_tools/generate-pages.js).
```
tidoust: That matches Webdex indeed, but it seems just as easy to include all `dfns`, so let's try to do that. This would also provide useful data to analyze: private terms should not be referenced by external specs in theory, the existence of such references usually signals that some discussion needs to take place between the groups responsible for the underlying specs.

test/backrefs/all.js
```diff
@@ -0,0 +1,109 @@
...
+const backrefsFolder = path.join(curatedFolder, 'backrefs');
+
+describe('The curated view of backrefs extracts', async () => {
+  const index = await loadJSON(path.join(curatedFolder, 'index.json'));
+  assert.ok(index, 'curated/index.json must exist (run npm run curate first)');
+
+  const shortnames = new Set(index.results.map(s => s.shortname));
+  const specsWithBackrefs = index.results.filter(s => s.backrefs);
+
+  it('lists backrefs extracts only for specs that have a backrefs file', async () => {
```
tidoust: That test should be dropped. It is useless and does not do what its label suggests it does in any case.

docs/backrefs-proposal.md
```diff
@@ -0,0 +1,45 @@
...
+
+```json
+{
+  "spec": { "title": "Streams Standard", "url": "https://streams.spec.whatwg.org/" },
+  "backrefs": [
+    {
+      "id": "readablestream",
+      "href": "https://streams.spec.whatwg.org/#readablestream",
+      "linkingText": ["ReadableStream"],
+      "type": "interface",
```
tidoust: So, one question here is how much information we want to copy from the `dfns` extract.

In theory, the `href` property is all we need, and people can then map the entry to the relevant `dfns` extract. If we end up publishing an npm package, it could perhaps be an extended version of the `@webref/xref` package that has both the `dfns` extracts and the `backrefs` ones, with simple JS logic to bind them.

But then if we want the backrefs extracts to be useful on their own, perhaps that's not the right approach. I need to think more about it. The decision could perhaps depend on the size of the generated data as well (the `dfns` extracts are already "big").

If we want the information to be more standalone, the `for` scoping property should also be copied over.

docs/backrefs-proposal.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+### Algorithm
+
+1. Index public dfns by `href` (plus HTML/ES multipage ↔ single-page aliases).
+2. Expand each referring spec’s `links` (`rawlinks` + `autolinks`) into `url#anchor` values.
+3. Match fragment URLs to dfns; accumulate referrers per defining term.
+4. Write one file per defining spec that has at least one referenced term; point `index.json` at it.
+
+Feedback welcome on folder name, `referencedBy` payload, and whether this should eventually move into Reffy as a `post` module.
```
tidoust: I don't have better suggestions for now ;)

tidoust (commented):
Many thanks for the PR.

I started to provide comments inline. I haven't reviewed the code yet. That will take more time. The overall approach seems sound and aligned with the way we usually "do things" in Webref!

docs/backrefs-proposal.md
DevOlabode: Done in the latest commit: backrefs now index all dfns (including private and argument definitions), and copy `access` so external references to private terms are visible for analysis.

test/backrefs/all.js
DevOlabode: Dropped that test.

docs/backrefs-proposal.md
DevOlabode: Went with the more standalone shape for now: each entry copies `id`, `href`, `linkingText`, `type`, `for`, and `access` from the dfns extract. Documented under README "Curation guarantees → Backrefs extracts". Happy to slim this down later if size or the npm-package direction suggests `href`-only is enough.

docs/backrefs-proposal.md
DevOlabode: Keeping `backrefs` and the current `referencedBy` payload for now then. Thanks!

DevOlabode:
Thanks for the review. I’ve addressed the open points:

- Dropped `docs/backrefs-proposal.md` and documented the backrefs schema/guarantees under **Curation guarantees** in the README
- Included **all** dfns (including private/argument), and copied `for` / `access` for standalone use
- Removed the useless backrefs file-existence test
- Fixed CI schema failures by skipping extract types that don’t have a Reffy JSON schema yet (`backrefs`)

All review threads should be resolved now. Happy to adjust further if needed.

tidoust (changes requested):
The code all looks good to me with a minor tweak: I would prefer to keep the same document order for the terms in the `backrefs` extracts and in the `dfns` extracts (dropping the call to `.sort` is enough to do that).

Now, thinking about it again, I don't think that we should integrate the code in Webref after all. We should rather do as for the other bits of processing and create a dedicated post-processing module in Reffy that Webref leverages in `prepare-curated.js`.

I prepared a draft pull request against Reffy that just does that, see https://github.com/w3c/reffy/pull/2127. The PR still needs to be completed with a JSON schema (and ideally a few tests), but the post-processing module seems to work correctly. Once Reffy is updated, this PR should:

1. drop `prepare-backref.js`
2. reuse the initial `prepare-curated.js` script, simply completing the list of post-processing modules in the call to `crawlSpecs` with `"backrefs"`
3. restore the `test/schemas.js` file (no need for an exception to the rule anymore).

I'm happy to be the one completing the PR in Reffy, but you're also welcome to take over and re-create the PR yourself (which would also have the advantage of making sure the update gets correctly associated with your GitHub profile). Up to you!

DevOlabode:
Thanks for the review and for suggesting the architectural change.

I've moved the backrefs generation into a dedicated Reffy post-processing module in **w3c/reffy#2128**, following the approach outlined in **w3c/reffy#2127**.

tidoust (changes requested):
The post-processing module is now in Reffy. This PR now needs to be updated along the lines suggested above (or a new PR needs to be created if that seems easier).

I'm not sure that much needs to remain in the `test/backrefs/all.js`, as tests in Reffy cover them already. But then it's always good to check again that the guarantees hold from a Webref perspective.

As before, I'm happy to take over. But equally happy to let you handle it! :)

DevOlabode:
Thanks! Since the Reffy work is now in place, I'm happy for you to take over the remaining Webref changes.

PR #2017, opened 2026-08-18
Add backrefs extracts per spec during curation

tidoust:
This leverages the new backrefs post-processing module exposed by Reffy to create backrefs extracts per spec. The update reuses content prepared by @DevOladobe in #1997.

The README now mentions the backrefs extracts as well, along with a list of guarantees, enforced by the tests.

Note that I also used this opportunity to link back to the relevant JSON schema that all curated data files are guaranteed to follow. For some reason, that wasn't specified anywhere.

This will produce ~10.4MB of data in 305 new extracts in a `backrefs` folder.

Tracking issue: #1934.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ More often than not, released versions of specifications are much older than the
 
 The following subfolders in the `curated` branch contain individual machine-readable JSON or text files generated from specifications:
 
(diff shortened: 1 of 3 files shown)


Elchi3 (approved):
Thanks @tidoust, looks reasonable to me.

PR #2040, opened 2026-09-04
Prefer `css-sizing` over `css-logical` to resolve duplicates

tidoust:
Some properties defined in `css-logical-1` got duplicated in `css-sizing-3`, see: https://github.com/w3c/csswg-drafts/issues/10189

This makes curation logic prefer `css-sizing-3` definitions so that curation correctly drops duplicate definitions.

Note: This should fix the curation step. CI tests will still fail due to a couple of hiccups in the events space. That's orthogonal to the problem at hand.


tidoust:
> Note: This should fix the curation step. CI tests will still fail due to a couple of hiccups in the events space. That's orthogonal to the problem at hand.

Well, no, CI tests no longer fail in practice. The hiccups were due to Specref being down but ReSpec now uses a fallback, so everything is fine.

